PR #24, opened 2023-10-26
feat: add database init Cloud Run job

averikitsch:


kurtisvg (changes requested):
I'm worried about this abstracting too much of the database set up away -- it feels like we are already pretty close with "run_database_init.py", I'm concerned that this PR pushes it over the edge into "too much abstraction"

averikitsch:
> I'm worried about this abstracting too much of the database set up away -- it feels like we are already pretty close with "run_database_init.py", I'm concerned that this PR pushes it over the edge into "too much abstraction"

Yes, this would aim at a demo quickstart rather than teaching users. I am fine with not maintaining this user journey but we may want to think about users be able to ssh into a compute instance.

PR #25, opened 2023-10-26
ci(extension): add integration tests for AlloyDB

averikitsch:


extension_service/alloydb.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,93 @@
...
+    name: python:3.11
+    dir: extension_service
+    entrypoint: pip
+    args: ["install", "-r", "requirements.txt", "-r", "requirements-test.txt", "--user"]
+
+  - id: Create database
+    name: postgres
+    entrypoint: /bin/bash
+    env:
+      - "PGUSER=${_DATABASE_USER}"
```
kurtisvg, line 26: nit: probably worthwhile to put this in a user too

extension_service/alloydb.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,93 @@
...
+    secretEnv:
+      - PGPASSWORD
+    args:
+      - '-c'
+      - |
+        psql -h ${_DATABASE_HOST} -c "CREATE DATABASE ${_DATABASE_NAME};" || echo "Already created."
+        psql -h ${_DATABASE_HOST} -d ${_DATABASE_NAME} -c 'CREATE EXTENSION vector;' || echo "Already created."
```
kurtisvg, line 33: nit: Can these two "Already created" output the name for easier debugging?

e.g. `"Database 'my-db' already created"`

extension_service/alloydb.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+  - id: Init database
```
kurtisvg, line 35:
```suggestion
  - id: Initialize data
```

kurtisvg: maybe nit -- does this need to be a separate step? looks like it's the same container as the previous step.

extension_service/alloydb.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,93 @@
...
+      - "DB_NAME=${_DATABASE_NAME}"
+      - "DB_USER=${_DATABASE_USER}"
+      - "DB_HOST=${_DATABASE_HOST}"
+    secretEnv:
+      - PGPASSWORD
+    entrypoint: /bin/bash
+    args:
+      - '-c'
+      - |
+        export DB_PASS=$$PGPASSWORD
```
kurtisvg, line 64: double checking -- is `$$` right?

extension_service/example-config.yml
```diff
@@ -3,8 +3,8 @@ host: 0.0.0.0
 datastore:
   # Example for AlloyDB
   kind: "postgres"
-  host: 127.0.0.1
-  # port: 5432
-  database: "my_database"
-  user: "my-user"
-  password: "my-password"
\ No newline at end of file
+  host: DB_HOST
```
kurtisvg, line 6: Does this hurt readability of this config? I wonder if we should just find/replace around it

kurtisvg (changes requested):

extension_service/alloydb.tests.cloudbuild.yaml
averikitsch: Yes, this is the format for env vars that aren't substitutions

extension_service/alloydb.tests.cloudbuild.yaml
averikitsch: What do you mean?

extension_service/alloydb.tests.cloudbuild.yaml
averikitsch: This could be combined with the test step but it has better debugging  to keep it separate with no noticeable increase in time

extension_service/alloydb.tests.cloudbuild.yaml
kurtisvg: I meant to say "in a secret"

kurtisvg (approved):

PR #29, opened 2023-10-27
ci: add integration tests for langchain-tools-demo

averikitsch:


langchain_tools_demo/int.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,54 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+steps:
+  - id: "Deploy to Cloud Run"
+    name: "gcr.io/cloud-builders/gcloud:latest"
+    dir: langchain_tools_demo
```
kurtisvg, line 17: Do we need to deploy the extension service as well?

langchain_tools_demo/int.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+serviceAccount: "projects/$PROJECT_ID/serviceAccounts/548341735270-compute@developer.gserviceaccount.com"
+options:
+  logging: CLOUD_LOGGING_ONLY
+  dynamic_substitutions: true
+
+substitutions:
+  _GCR_HOSTNAME: ${_REGION}-docker.pkg.dev
+  _SERVICE: demo-service-${BUILD_ID}
+  _REGION: us-central1
```
kurtisvg, line 54: Should these be at the top?

kurtisvg (commented):
I'm confused on how this is being triggered. Is something configure outside of the repo? If so can we bring it into the repo? if not, can we document instructions for accessing it and setting it up again?

langchain_tools_demo/int.tests.cloudbuild.yaml
averikitsch: They can be listed at the top or bottom. I prefer bottom.

langchain_tools_demo/int.tests.cloudbuild.yaml
averikitsch: Instead of writing unit tests for the demo, the MVtesting is the build and deploy seen here. The extension service is not needed to test the demo deployment. The demo can handle errors when trying to contact the extension. We could update this in the future to be the full E2E test for both services

DEVELOPER.md
```diff
@@ -43,34 +42,30 @@ pip install -r extension_service/requirements-test.txt -r langchain_tools_demo/r
...
+1. To use a live extension service on Cloud Run:
 
-    ```bash
-    export PORT=9090
-    ```
+    1. Set up [Application Default Credentials][ADC]:
 
-1. [Optional] Set `BASE_URL` environment variable:
+        ```bash
+        export GOOGLE_APPLICATION_CREDENTIALS=/path/to/key.json
```
kurtisvg, line 56: We should encourage `gcloud auth application-default login` over a service account key

kurtisvg (approved):

PR #34, opened 2023-10-30
feat: Update prompt and tools

averikitsch:


kurtisvg (approved):

langchain_tools_demo/tools.py
```diff
@@ -0,0 +1,256 @@
...
+    )
+    arrival_airport: Optional[str] = Field(description="Arrival airport 3-letter code")
+    date: Optional[str] = Field(description="Date of flight departure")
+
+
+@tool("List Flights", args_schema=ListFlights)
+def list_flights(departure_airport: str, arrival_airport: str, date: str):
+    """
+    Use this tool to list all flights matching search criteria.
+    Takes an arrival airport and departure airport and returns all flights.
```
duwenxin99, line 156: This should include the date as well? eg. "Takes an arrival airport, a departure airport, or both, filters by date and returns all matching flights."

langchain_tools_demo/tools.py
```diff
@@ -0,0 +1,256 @@
...
+    The agent can decide to return the results directly to the user.
+    Input of this tool must be in JSON format and include all three inputs - arrival_airport, departure_airport, and date.
```
duwenxin99, line 158: The search flight endpoint supports search with only `arrival_airport` or only `departure_airport` and doesn't require both. Do we want to disallow that for the agent? The user may want to search for "all flights leaving SFO today".

langchain_tools_demo/tools.py
```diff
@@ -0,0 +1,256 @@
...
+
+    return response.json()
+
+
+class QueryInput(BaseModel):
+    query: str = Field(description="Search query")
+
+
+@tool("Search Amenities", args_schema=QueryInput)
+def search_amenities(query: str):
```
duwenxin99, line 229: Maybe we should call this tool `list_amenities` to match `list_flights`?

extension_service/app/routes.py
```diff
@@ -65,7 +65,18 @@ async def search_flights(
...
     departure_airport: Optional[str] = None,
     arrival_airport: Optional[str] = None,
+    date: Optional[str] = None,
+    airline: Optional[str] = None,
+    flight_number: Optional[str] = None,
 ):
     ds: datastore.Client = request.app.state.datastore
-    flights = await ds.search_flights(departure_airport, arrival_airport)
+    if date and (arrival_airport or departure_airport):
+        flights = await ds.search_flights(date, departure_airport, arrival_airport)
```
Yuan325, line 74:
```suggestion
        flights = await ds.search_flights_by_airports(date, departure_airport, arrival_airport)
```

extension_service/datastore/datastore.py
```diff
@@ -77,9 +77,18 @@ async def amenities_search(
...
+    @abstractmethod
+    async def search_flights_by_number(
+        self,
+        airline: str,
+        flight_number: str,
+    ) -> Optional[list[models.Flight]]:
+        raise NotImplementedError("Subclass should implement this!")
+
     @abstractmethod
     async def search_flights(
```
Yuan325, line 89:
```suggestion
    async def search_flights_by_airports(
```

extension_service/datastore/providers/postgres.py
```diff
@@ -251,8 +252,27 @@ async def get_flight(self, flight_id: int) -> Optional[list[models.Flight]]:
...
+                AND flight_number = $2;
+            """,
+            airline,
+            number,
+            timeout=10,
+        )
+        flights = [models.Flight.model_validate(dict(r)) for r in results]
+        return flights
+
     async def search_flights(
```
Yuan325, line 273:
```suggestion
    async def search_flights_by_airports(
```

Yuan325 (changes requested):
nit: update search_flights to search_flights_by_airports to make it consistent. suggested edit to all fields that need changes.

langchain_tools_demo/tools.py
averikitsch: I like search amenities because it's using the vector similarity search, but I will leave this open for future discussion

langchain_tools_demo/tools.py
averikitsch: This aims at getting the agent to include all three inputs even if they are null. We would have to subclass BaseTool in order to fix the validation if the input, even though optional, isn't included.

langchain_tools_demo/tools.py
averikitsch: I will update this, but from experiments I saw that the tool's args_schema is more important than the "prompt"/tool description but the tool description can definitely help the tool to be used correctly.

duwenxin99 (approved):

PR #38, opened 2023-11-01
chore: refactor get airport endpoint and add search airport endpoint

Yuan325:
Refactor get airport endpoint to include iata as param. 
Add search airports endpoint with city and/or country and/or name as param.

Did not add unit test for the providers now, will work on tests when add integration test in future PR.

### DIFF
--- a/extension_service/datastore/datastore.py
+++ b/extension_service/datastore/datastore.py
@@ -60,7 +60,20 @@ class Client(ABC, Generic[C]):
         pass
 
     @abstractmethod
-    async def get_airport(self, id: int) -> Optional[models.Airport]:
+    async def get_airport_by_id(self, id: int) -> Optional[models.Airport]:
+        raise NotImplementedError("Subclass should implement this!")
+
+    @abstractmethod
+    async def get_airport_by_iata(self, iata: str) -> Optional[models.Airport]:
+        raise NotImplementedError("Subclass should implement this!")
+
+    @abstractmethod
+    async def search_airports(
+        self,
+        country: Optional[str] = None,
+        city: Optional[str] = None,
+        name: Optional[str] = None,
+    ) -> Optional[list[models.Airport]]:
         raise NotImplementedError("Subclass should implement this!")
 
     @abstractmethod
--- a/extension_service/datastore/datastore.py
+++ b/extension_service/datastore/datastore.py
@@ -73,7 +73,7 @@ class Client(ABC, Generic[C]):
         country: Optional[str] = None,
         city: Optional[str] = None,
         name: Optional[str] = None,
-    ) -> Optional[list[models.Airport]]:
+    ) -> list[models.Airport]:
         raise NotImplementedError("Subclass should implement this!")
 
     @abstractmethod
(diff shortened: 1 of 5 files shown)


extension_service/datastore/providers/postgres.py
```diff
@@ -184,12 +184,25 @@ async def export_data(
...
+        name: Optional[str] = None,
+    ) -> Optional[models.Airport]:
+        id_str = "%" if id is None else str(id)
+        iata = "%" if iata is None else iata
+        name = "%" if name is None else "%" + name + "%"
+
         result = await self.__pool.fetchrow(
             """
-              SELECT id, iata, name, city, country FROM airports WHERE id=$1
+              SELECT * FROM airports WHERE CAST(id AS text) LIKE $1
```
averikitsch, line 199: Could this query since it's using the like operator potentially return multiple results?

Yuan325: Updated the function. Created 2 separate function and move name under search_airport.

extension_service/datastore/providers/postgres.py
```diff
@@ -198,6 +198,52 @@ async def get_airport(self, id: int) -> Optional[models.Airport]:
...
+        results = await self.__pool.fetch(
+            """
+            SELECT * FROM (
+                SELECT * FROM (
+                    SELECT * FROM airports
+                    WHERE country ILIKE $1
+                ) AS filtered_country
+                WHERE city ILIKE $2
+            ) AS filtered_city
+            WHERE name ILIKE $3
```
kurtisvg, line 236: This seemed like too much nesting, so I asked Bard and it recommend this:

```
SELECT * FROM users
WHERE ($1 IS NULL OR country ILIKE $1)
AND ($2 NULL OR city ILIKE $2)
AND ($3 IS NULL OR name LIKE '%$3%')
```

(Which seems pretty good to me if it works)

kurtisvg (changes requested):

extension_service/datastore/datastore.py
```diff
@@ -60,7 +60,20 @@ async def export_data(
...
+    async def get_airport_by_iata(self, iata: str) -> Optional[models.Airport]:
+        raise NotImplementedError("Subclass should implement this!")
+
+    @abstractmethod
+    async def search_airports(
+        self,
+        country: Optional[str] = None,
+        city: Optional[str] = None,
+        name: Optional[str] = None,
+    ) -> Optional[list[models.Airport]]:
```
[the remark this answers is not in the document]
Yuan325: Removed the Optional.
It was previously Optional to keep it consistent with the Get endpoints (if no airport/amenities was found, it will return None). But I think it does make sense to return empty list for search endpoints. Updated.

extension_service/datastore/providers/postgres.py
```diff
@@ -199,6 +199,48 @@ async def get_airport(self, id: int) -> Optional[models.Airport]:
...
+        if city is None:
+            city = "%"
+        name = "%" if name is None else f"%{name}%"
+
+        results = await self.__pool.fetch(
+            """
+            SELECT * FROM airports
+            WHERE (country IS NULL OR country ILIKE $1)
+            AND (city IS NULL OR city ILIKE $2)
+            AND (name IS NULL OR name ILIKE $3)
```
kurtisvg, line 233: I think this should be `($1 IS NULL OR country ILIKE $1)` -- essentially skip the statement if not provided. Then you can also remove `if country is None` section earlier.

kurtisvg (approved):

extension_service/datastore/providers/postgres.py
Yuan325: Updated! Thank youuu

PR #42, opened 2023-11-07
feat: Add Renovate Config

duwenxin99:


.github/renovate.extension.json
```diff
@@ -0,0 +1,36 @@
...
+    "commitMessagePrefix": "deps: ",
+    "ignorePresets": [":semanticPrefixFixDepsChoreOthers"],
+    "ignorePaths": ["docs/", "data/"],
+    "prConcurrentLimit": 0,
+    "rebaseWhen": "behind-base-branch",
+    "dependencyDashboard": true,
+    "dependencyDashboardLabels": ["type: process"],
+    "semanticCommits": "disabled",
+    "pip_requirements": {
+      "fileMatch": ["extension_service/requirements.txt", "extension_service/requirements-test.txt"]
```
kurtisvg, line 16: Shouldn't this cover the github repo as well?

also super-nit: prefer to use a multi line list here to make it easier to add/remove in the future

.github/renovate.extension.json
```diff
@@ -0,0 +1,36 @@
...
+    },
+    "packageRules": [
+      {
+        "matchPackagePatterns": [
```
kurtisvg, line 20: Do we really need to group these? What's the rationale?

kurtisvg (changes requested):

.github/renovate.extension.json
duwenxin99: deleted the grouping for github dependencies.

.github/renovate.json
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    "commitMessagePrefix": "deps: ",
+    "ignorePresets": [
+        ":semanticPrefixFixDepsChoreOthers"
+    ],
+    "ignorePaths": [
+        "docs/",
+        "data/",
+        ".github/"
+    ],
```
averikitsch, line 15: nit: not needed

.github/renovate.json
```diff
@@ -0,0 +1,29 @@
...
+    "prConcurrentLimit": 0,
+    "rebaseWhen": "behind-base-branch",
+    "dependencyDashboard": true,
+    "dependencyDashboardLabels": [
+        "type: process"
+    ],
+    "semanticCommits": "disabled",
```
averikitsch, line 22: This seems in conflict with the other semantic commit settings

.github/renovate.json
```diff
@@ -0,0 +1,29 @@
...
+    "pip_requirements": {
+        "fileMatch": [
+            "*.requirements.txt",
+            "*.requirements-test.txt"
```
averikitsch, line 26: Only `requirements-test.txt` should be needed, since the requirements.txt is added by default. It seems like the regex `*.` might not be needed, see https://docs.renovatebot.com/modules/manager/pip_requirements/

.github/renovate.json
duwenxin99: deleted

.github/renovate.json
duwenxin99: deleted regex expression

.github/renovate.json
kurtisvg: My memory is rust, but I think this is because the other one is doing "chore" and this one uses "fix" or something by default...

kurtisvg (approved):

.github/renovate.json
duwenxin99: I had some confusions over the default semantic commits behavior. The default value is `auto` which auto-detects if the repo had semantic commits messages. Seems like `"semanticCommits": "disabled"` would disable the semantic commits entirely, so I changed it to `enabled` so all the commits WILL be semantic commits. ps: I think Kurtis is referring to `":semanticCommitTypeAll(chore)"` which prefixes all commits with `chore`; without it the default uses both `chore` and `fix`, but seems like the `semanticCommits` still needs to be enabled for that to work...

kurtisvg: Renovate is kinda iffy on config stuff sometimes -- sometimes the default configs change and it's hard to understand why. Looking at these two docs:
https://docs.renovatebot.com/semantic-commits/
https://docs.renovatebot.com/presets-default/#semanticcommits

I would say that we can use either -- so let's just pick your favorite and if it doesn't work we'll PR it to the other one.

averikitsch (approved):

PR #65, opened 2023-11-14
feat: Add Firestore provider

duwenxin99:
Data models: [Database Query Extension Data Models](https://docs.google.com/document/d/1_VJnMu31xCG1Y3DeLiAInPGyDJKYKgpXHfSFgndVqak/edit?resourcekey=0-QWFpGiz2zfkGbseG-gfXEA#heading=h.u85htyfahygy)

### DIFF
--- a/extension_service/datastore/__init__.py
+++ b/extension_service/datastore/__init__.py
@@ -17,7 +17,6 @@ from typing import Union
 from . import providers
 from .datastore import Client, create
 
-# Temporarily list both types twice until we have more kinds
 Config = Union[providers.firestore.Config, providers.postgres.Config]
 
 __ALL__ = [Client, Config, create, providers]
--- a/extension_service/datastore/__init__.py
+++ b/extension_service/datastore/__init__.py
@@ -18,6 +18,6 @@ from . import providers
 from .datastore import Client, create
 
 # Temporarily list both types twice until we have more kinds
-Config = Union[providers.postgres.Config, providers.postgres.Config]
+Config = Union[providers.firestore.Config, providers.postgres.Config]
 
 __ALL__ = [Client, Config, create, providers]
--- a/extension_service/datastore/providers/firestore.py
+++ b/extension_service/datastore/providers/firestore.py
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import Any, Dict, Literal, Optional
 
-import firebase_admin  # type: ignore
+import firebase_admin
 from firebase_admin import credentials, firestore_async
 from google.cloud import firestore
 from google.cloud.firestore_v1.base_query import FieldFilter
@@ -204,7 +204,7 @@ class Client(datastore.Client[Config]):
 
     async def amenities_search(
         self, query_embedding: list[float], similarity_threshold: float, top_k: int
-    ) -> Optional[list[models.Amenity]]:
+    ) -> list[models.Amenity]:
         query = (
             self.__client.collection("amenities")
             .where("embedding", ">", 1 - similarity_threshold)
@@ -220,7 +220,7 @@ class Client(datastore.Client[Config]):
         amenities = [models.Amenity.model_validate(dict(doc)) for doc in docs]
         return amenities
 
-    async def get_flight(self, flight_id: int) -> Optional[list[models.Flight]]:
+    async def get_flight(self, flight_id: int) -> Optional[models.Flight]:
         query = self.__client.collection("flights").where(
             filter=FieldFilter("id", "==", id)
         )
@@ -230,7 +230,7 @@ class Client(datastore.Client[Config]):
         self,
         airline: str,
         number: str,
-    ) -> Optional[list[models.Flight]]:
+    ) -> list[models.Flight]:
         query = (
             self.__client.collection("flights")
             .where(filter=FieldFilter("airline", "==", airline))
@@ -247,14 +247,14 @@ class Client(datastore.Client[Config]):
         date: str,
         departure_airport: Optional[str] = None,
         arrival_airport: Optional[str] = None,
-    ) -> Optional[list[models.Flight]]:
+    ) -> list[models.Flight]:
         # Check if either parameter is null.
-
+        date = datetime.strptime(date, "%Y-%m-%d").date()
         date_timestamp = datetime.combine(date, datetime.min.time())
         query = (
             self.__client.collection("flights")
-            .where("departure_time", ">=", date_timestamp - datetime.timedelta(days=1))
-            .where("departure_time", "<=", date_timestamp + datetime.timedelta(days=1))
+            .where("departure_time", ">=", date_timestamp - timedelta(days=1))
+            .where("departure_time", "<=", date_timestamp + timedelta(days=1))
         )
 
         if departure_airport is None:
--- a/extension_service/datastore/providers/firestore.py
+++ b/extension_service/datastore/providers/firestore.py
@@ -59,39 +59,39 @@ class Client(datastore.Client[Config]):
         amenities: list[models.Amenity],
         flights: list[models.Flight],
     ) -> None:
-        for a in airports:
-            await self.__client.collection("airports").document(str(a.id)).set(
+        for airport in airports:
+            await self.__client.collection("airports").document(str(airport.id)).set(
                 {
-                    "iata": a.iata,
-                    "name": a.name,
-                    "city": a.city,
-                    "country": a.country,
+                    "iata": airport.iata,
+                    "name": airport.name,
+                    "city": airport.city,
+                    "country": airport.country,
                 }
             )
-        for a in amenities:
-            await self.__client.collection("amenities").document(str(a.id)).set(
+        for amenity in amenities:
+            await self.__client.collection("amenities").document(str(amenity.id)).set(
                 {
-                    "name": a.name,
-                    "description": a.description,
-                    "location": a.location,
-                    "terminal": a.terminal,
-                    "category": a.category,
-                    "hour": a.hour,
-                    "content": a.content,
-                    "embedding": a.embedding,
+                    "name": amenity.name,
+                    "description": amenity.description,
+                    "location": amenity.location,
+                    "terminal": amenity.terminal,
+                    "category": amenity.category,
+                    "hour": amenity.hour,
+                    "content": amenity.content,
+                    "embedding": amenity.embedding,
                 }
             )
-        for f in flights:
-            await self.__client.collection("flights").document(str(f.id)).set(
+        for flight in flights:
+            await self.__client.collection("flights").document(str(flight.id)).set(
                 {
-                    "airline": f.airline,
-                    "flight_number": f.flight_number,
-                    "departure_airport": f.departure_airport,
-                    "arrival_airport": f.arrival_airport,
-                    "departure_time": f.departure_time,
-                    "arrival_time": f.arrival_time,
-                    "departure_gate": f.departure_gate,
-                    "arrival_gate": f.arrival_gate,
+                    "airline": flight.airline,
+                    "flight_number": flight.flight_number,
+                    "departure_airport": flight.departure_airport,
+                    "arrival_airport": flight.arrival_airport,
+                    "departure_time": flight.departure_time,
+                    "arrival_time": flight.arrival_time,
+                    "departure_gate": flight.departure_gate,
+                    "arrival_gate": flight.arrival_gate,
                 }
             )
 
--- a/extension_service/datastore/providers/firestore.py
+++ b/extension_service/datastore/providers/firestore.py
@@ -105,19 +105,17 @@ class Client(datastore.Client[Config]):
     async def export_data(
         self,
     ) -> tuple[list[models.Airport], list[models.Amenity], list[models.Flight]]:
-        airport_docs = self.__client.collection("airports").stream()
-        amenities_docs = self.__client.collection("amenities").stream()
-        flights_docs = self.__client.collection("flights").stream()
+        airport_docs = await self.__client.collection("airports").get()
+        amenities_docs = await self.__client.collection("amenities").get()
+        flights_docs = await self.__client.collection("flights").get()
 
         airports = [
-            models.Airport.model_validate(doc.to_dict()) for doc in await airport_docs
+            models.Airport.model_validate(doc.to_dict()) for doc in airport_docs
         ]
         amenities = [
-            models.Amenity.model_validate(doc.to_dict()) for doc in await amenities_docs
+            models.Amenity.model_validate(doc.to_dict()) for doc in amenities_docs
         ]
-        flights = [
-            models.Flight.model_validate(doc.to_dict()) for doc in await flights_docs
-        ]
+        flights = [models.Flight.model_validate(doc.to_dict()) for doc in flights_docs]
         return airports, amenities, flights
 
     async def get_airport(self, id: int) -> Optional[models.Airport]:
@@ -159,11 +157,11 @@ class Client(datastore.Client[Config]):
             .limit(top_k)
         )
 
-        docs = query.stream()
+        docs = await query.get()
         if docs is []:
             return None
 
-        amenities = [models.Amenity.model_validate(dict(doc)) async for doc in docs]
+        amenities = [models.Amenity.model_validate(dict(doc)) for doc in docs]
         return amenities
 
     async def get_flight(self, flight_id: int) -> Optional[list[models.Flight]]:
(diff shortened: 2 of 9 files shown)


extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+    @classmethod
+    async def create(cls, config: Config) -> "Client":
+        cred = credentials.ApplicationDefault()
+        firebase_admin.initialize_app(
+            cred,
+            options={
+                "projectId": config.projectId,
+                "serviceAccountId": config.serviceAccountId,
+            },
+        )
```
kurtisvg, line 54: Are you sure we need to do this? Is this creating a new app or just representing the App locally?

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+import models
+
+from .. import datastore
+
+
+class Config(BaseModel, datastore.AbstractConfig):
+    kind: Literal["firestore"]
+    projectId: str
+    serviceAccountId: str
+
```
kurtisvg, line 33: 1. I think we can let projectId be optional -- if it's unset we shouldn't pass into the client and just let it use the default
2. I don't think we need service account id

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+        async def delete_collection(coll_ref, batch_size=400):
+            docs = coll_ref.limit(batch_size).stream()
+            deleted = 0
+
+            async for doc in docs:
+                await doc.reference.delete()
+                deleted = deleted + 1
+
+            if deleted >= batch_size:
+                return await delete_collection(coll_ref, batch_size)
```
kurtisvg, line 72: I don't understand what the batch_size is or why it's set to 400 (or why it's a parameter). Why are we counting and returning delete_collection here? Can we include some comments on why this is necessary?

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+
+        amenities_ref = self.__client.collection("amenities")
+        amenities_exist = await amenities_ref.limit(1).get()
+        if amenities_exist:
+            await delete_collection(amenities_ref)
+
+        flights_ref = self.__client.collection("flights")
+        flights_exist = await flights_ref.limit(1).get()
+        if flights_exist:
+            await delete_collection(flights_ref)
```
kurtisvg, line 88: Any reason we shouldn't try to await these together -- and get some benefit from async here?

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+                    "airline": flight.airline,
+                    "flight_number": flight.flight_number,
+                    "departure_airport": flight.departure_airport,
+                    "arrival_airport": flight.arrival_airport,
+                    "departure_time": flight.departure_time,
+                    "arrival_time": flight.arrival_time,
+                    "departure_gate": flight.departure_gate,
+                    "arrival_gate": flight.arrival_gate,
+                }
+            )
```
kurtisvg, line 125: This seems like another good target for asyncronization

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+        airport_docs = self.__client.collection("airports").stream()
+        amenities_docs = self.__client.collection("amenities").stream()
+        flights_docs = self.__client.collection("flights").stream()
+
+        airports, amenities, flights = [], [], []
+
+        for doc in airport_docs:
+            airport_dict = doc.to_dict()
+            airport_dict["id"] = doc.id
+            airports.append(models.Airport.model_validate(airport_dict))
```
kurtisvg, line 139: nit: it's best practice to limit the scope of the variables to when you need them:

```suggestion
        airports = []
        for doc in airport_docs:
            airport_dict = doc.to_dict()
            airport_dict["id"] = doc.id
            airports.append(models.Airport.model_validate(airport_dict))
            
        amenities = []
        # ...
```

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+            flight_dict["id"] = doc.id
+            flights.append(models.Flight.model_validate(flight_dict))
+
+        return airports, amenities, flights
+
+    async def get_airport_by_id(self, id: int) -> Optional[models.Airport]:
+        query = self.__client.collection("airports").where(
+            filter=FieldFilter("id", "==", id)
+        )
+        return models.Airport.model_validate(query.stream().to_dict())
```
kurtisvg, line 157: Does query need to be awaited somehow?

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+
+    async def search_airports(
+        self,
+        country: Optional[str] = None,
+        city: Optional[str] = None,
+        name: Optional[str] = None,
+    ) -> list[models.Airport]:
+        query = self.__client.collection("airports")
+
+        if country:
```
kurtisvg, line 173: maybe want to make this `if country is not None` to avoid issues with falsy values

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+                "id",
+                "name",
+                "description",
+                "location",
+                "terminal",
+                "category",
+                "hour",
+                "content",
+                "embedding",
+            )
```
kurtisvg, line 201: Do we need to specify select here? Not sure we do it in other places

extension_service/datastore/providers/firestore_test.py
```diff
@@ -0,0 +1,317 @@
...
+
+from google.cloud import firestore
+from google.cloud.firestore_v1.base_query import FieldFilter
+
+import models
+
+from . import firestore as firestore_provider
+
+
+class MockDocument(Dict):
```
kurtisvg, line 26: I think maybe we don't want to use Mock's here, since we are trying to move away from Mocks in Postgres. Maybe @Yuan325 can weigh in on what she'd prefer?

kurtisvg (changes requested):
Notes: 
 - We need a documentation page for this (something like what we have for AlloyDB)

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,270 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from datetime import datetime, timedelta
+from typing import Any, Dict, Literal, Optional
+
+import firebase_admin
+from firebase_admin import credentials, firestore_async
```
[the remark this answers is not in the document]
duwenxin99: Kept only firebase_admin module import

extension_service/datastore/providers/firestore.py
duwenxin99: Made projectId optional. Deleted service account id field.

extension_service/datastore/providers/firestore.py
duwenxin99: When deleting a collection in firestore, we need to retrieve all documents and delete each one. It is recommended to delete them in smaller batches to avoid `out-of-memory` error according to https://firebase.google.com/docs/firestore/manage-data/delete-data#collections. Since a batched write can contain maximum 500 operations, `batch_size` should be < 500 and I set it to 400 to be safe. It doesn't have to be a parameter though, I could just use a hardcoded value if that looks better.

extension_service/datastore/providers/firestore.py
duwenxin99: Deleted

extension_service/datastore/providers/firestore.py
duwenxin99: added await

extension_service/datastore/providers/firestore.py
duwenxin99: fixed

extension_service/datastore/providers/firestore.py
duwenxin99: Previously `amenities_search` in postgres.py has selection clause so I matched its query. Seems like it is no longer the case so I will delete select clause here as well.

extension_service/datastore/providers/firestore_test.py
Yuan325: Yeah, most likely it will be re-written after the integration test is implemented.

duwenxin99: I'll leave it for now and re-write it later then.

extension_service/datastore/providers/firestore.py
duwenxin99: After discussion with Daniel, we determined it is unnecessary to delete recursively in batches. Instead we iteratively create task for each delete, and await them together with `asyncio.gather()`

extension_service/datastore/providers/firestore.py
duwenxin99: Updated: we will await all the delete together.

extension_service/datastore/providers/firestore.py
duwenxin99: Updated with async.

extension_service/datastore/__init__.py
```diff
@@ -18,6 +18,6 @@
 from .datastore import Client, create
 
 # Temporarily list both types twice until we have more kinds
-Config = Union[providers.postgres.Config, providers.postgres.Config]
+Config = Union[providers.firestore.Config, providers.postgres.Config]
```
kurtisvg, line 21: nit: can remove the comment about this now

extension_service/datastore/providers/firestore.py
kurtisvg: I don't see anything missing, so I'm guessing not.

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,281 @@
...
+            )
+        )
+        return models.Amenity.model_validate(query.stream().to_dict())
+
+    async def amenities_search(
+        self, query_embedding: list[float], similarity_threshold: float, top_k: int
+    ) -> list[models.Amenity]:
+        query = (
+            self.__client.collection("amenities")
+            .where("embedding", ">", 1 - similarity_threshold)
```
kurtisvg, line 218: I don't follow this -- does Firestore support Vector search? I don't think this is doing what we want it to do. 

This might be one of the gaps where we need to find an alternative or not use until we can update with semantic similarity search.

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,281 @@
...
+            )
+            .order_by("similarity", direction="descending")
+            .limit(top_k)
+        )
+
+        docs = query.stream()
+        amenities = []
+        async for doc in docs:
+            amenities.append(models.Amenity.model_validate(dict(doc)))
+        return amenities
```
kurtisvg, line 230: nit: can we use list comprehension here

extension_service/datastore/providers/firestore.py
kurtisvg: Ok, I keep seeing this and I don't understand it. 

Shouldn't we need to await the stream or call `.get` and await that? This example shows them awaiting from `.get()`: 
https://cloud.google.com/firestore/docs/query-data/get-data

This example shows them awaiting `.stream()`: https://cloud.google.com/firestore/docs/query-data/get-data#get_all_documents_in_a_collection

kurtisvg (changes requested):

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,281 @@
...
+        self,
+        airports: list[models.Airport],
+        amenities: list[models.Amenity],
+        flights: list[models.Flight],
+    ) -> None:
+        async def delete_collections(collection_list: list[AsyncCollectionReference]):
+            # Checks if colelction exists and deletes all documents
+            delete_tasks = []
+            for collection_ref in collection_list:
+                collection_exists = collection_ref.limit(1).stream()
```
averikitsch, line 58: nit: do we want a bigger batch size than 1?

extension_service/datastore/providers/firestore.py
duwenxin99: I don't see this example awating `.stream()`? It uses `async for` to loop the stream but does not await? When I tried to add await before `.stream()`, it gave me error `TypeError: object async_generator can't be used in 'await' expression`

duwenxin99: I think `.get()` needs to be awaited because it returns the document snapshots themselves, while `.stream()` only returns the generator object to be consumed

extension_service/datastore/providers/firestore.py
duwenxin99: We are only checking if the collection exists here, so selecting 1 document within the collection should be sufficient.

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,281 @@
...
+        docs = query.stream()
+        airports = []
+        async for doc in docs:
+            airports.append(models.Airport.model_validate(dict(doc)))
+        return airports
+
+    async def get_amenity(self, id: int) -> Optional[models.Amenity]:
+        query = (
+            self.__client.collection("amenities")
+            .where(filter=FieldFilter("id", "==", id))
```
averikitsch, line 198: What the difference between using FieldFilter vs where directly>

extension_service/datastore/providers/firestore.py
duwenxin99: Sorry this was just a place holder and I missed it...Maybe we can try using https://extensions.dev/extensions/googlecloud/firestore-semantic-search? This extension seem like a plausible alternative?

extension_service/datastore/providers/firestore.py
duwenxin99: Updated with list comprehension.

extension_service/datastore/providers/firestore.py
kurtisvg: I'm not sure how well that will work -- it looks like it's creating the embedding as part of the query and using Vertex Vector Search on the backend... we would need to set up Vector Search which is a lot more expensive than just firestore. 

My gut is we will just need to throw not implemented and figure out how to demonstrate that in the front end.

extension_service/datastore/providers/firestore.py
duwenxin99: Firestore doesn't mention any difference on their official documentation... But it does seem like most of their filter examples use `FieldFilter` instead of passing in directly. I guess it might be mostly for params passing purposes (so you could pass other params into the `where` method besides the filter).

extension_service/datastore/providers/firestore.py
duwenxin99: Raise Unimplemented Error for now.

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,249 @@
...
+        amenities: list[models.Amenity],
+        flights: list[models.Flight],
+    ) -> None:
+        async def delete_collections(collection_list: list[AsyncCollectionReference]):
+            # Checks if colelction exists and deletes all documents
+            delete_tasks = []
+            for collection_ref in collection_list:
+                collection_exists = collection_ref.limit(1).stream()
+                if not collection_exists:
+                    return
```
kurtisvg, line 60: nit: should this be `continue` over `return? I think we want to make sure all collections are deleted

extension_service/datastore/providers/firestore.py
```diff
@@ -0,0 +1,249 @@
...
+        number: str,
+    ) -> list[models.Flight]:
+        query = (
+            self.__client.collection("flights")
+            .where(filter=FieldFilter("airline", "==", airline))
+            .where(filter=FieldFilter("flight_number", "==", number))
+        )
+
+        docs = query.stream()
+        flights = [models.Flight.model_validate(dict(doc)) async for doc in docs]
```
kurtisvg, line 222: super-nit: 
```suggestion
        flights = [models.Flight.model_validate(dict(doc)) async for doc in query.stream()]
```

kurtisvg (approved):

extension_service/pyproject.toml
```diff
@@ -2,9 +2,9 @@
 profile = "black"
 
 [tool.mypy]
-python_version = 3.10
+python_version = "3.10"
 warn_unused_configs = true
 
 [[tool.mypy.overrides]]
-module = "pgvector.asyncpg"
+module = ["pgvector.asyncpg", "firebase_admin"]
```
kurtisvg, line 9: I think need need to remove firebase_admin here

extension_service/requirements.txt
```diff
@@ -1,5 +1,6 @@
 asyncpg==0.29.0
 fastapi==0.104.1
+firebase_admin==6.2.0
```
kurtisvg, line 3: I think this should be replaced with https://pypi.org/project/async-firebase/? (I think it's the same library you are using)

kurtisvg (approved):

extension_service/datastore/providers/firestore.py
duwenxin99: Updated with `return`

extension_service/requirements.txt
duwenxin99: This `async-firebase` is actually a super light-weight wrapper used for android and ios messaging only and is not the same as what we are using. Replaced `firebase-admin` with https://pypi.org/project/google-cloud-firestore/.

PR #94, opened 2023-11-21
fix(firestore): Add ID to all documents in Firestore provider

anihm136:
Noticed a bug in the Firestore provider while taking this on a test run - many of the methods with the Firestore provider do not add the ID to the document and therefore fail model validation. Additionally, some places had incorrect conversion of documents into dict (using `dict(doc)` instead of `doc.to_dict()`). This PR resolves both of these issues.


kurtisvg (approved):

duwenxin99 (approved):

anihm136:
Have pushed a fix for the failing test

kurtisvg:
Thanks @anihm136!

PR #99, opened 2023-11-23
feat: add clean up instructions

Yuan325:


docs/clean_up.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Delete Cloud Firestore resources
+
+1. Go to your [project settings][firebase-settings] in the Firebase console.
+
+1. At the bottom of the General settings page, click Delete project.
+
+1. Select each checkbox to acknowledge the effects of deleting your project.
+
+1. Click Delete project.
```
averikitsch, line 67: It looks like we cannot delete the default database? But we could delete a default database https://firebase.google.com/docs/firestore/manage-databases#delete_a_database. I believe that if you don't use the firestore db then you shouldn't be charged anything. We should make it clear that delete the project is more extreme than a clean up.

averikitsch (approved):
Approved, feel free to use or disregard comments

docs/clean_up.md
Yuan325: I did delete project since the firestore (docs/datastore/firestore.md) instructions asked user to create a project.

docs/clean_up.md
```diff
@@ -0,0 +1,69 @@
+# Clean up
+
```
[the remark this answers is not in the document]
Yuan325: Added a quick note regarding this.

kurtisvg (changes requested):

docs/clean_up.md
Yuan325: Yes, should I just remove the delete project instruction overall and replace with deleting the database? (Though our set up firestore instructions did ask user to create a new project). I was viewing clean up instructions as cleaning up things that we have user create when following our instruction.

README.md
```diff
@@ -132,6 +132,10 @@ of any of the database's listed below to set up and initialize your database:
 
 [Instructions for running app locally](./docs/run_langchain_demo.md)
 
+### Clean up resources
+
+[Instructions for cleaning up resources](./docs/clean_up.md)
```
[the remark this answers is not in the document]
Yuan325: Yes, the format is correct. Multiple clean up instructions might be too much since the clean up instructions is not too long. I can move the datastore specific clean up to their own setup docs (plus, mention to not clean up until the end of demo) and link it from the clean up doc.

docs/clean_up.md
kurtisvg: My 2c is that the project requirement should be consistent. It's probably a "before you begin thing" and so doesn't need to be part of the clean up instructions.

kurtisvg (approved):

PR #104, opened 2023-11-27
chores: update langchain python version requirement to 3.10

Yuan325:
Tested that 3.10 works.

### DIFF
--- a/langchain_tools_demo/Dockerfile
+++ b/langchain_tools_demo/Dockerfile
@@ -14,7 +14,7 @@
 
 # Use the official lightweight Python image.
 # https://hub.docker.com/_/python
-FROM python:3.11-slim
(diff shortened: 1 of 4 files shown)


kurtisvg:
What's the motivation to move backwards?

langchain_tools_demo/Dockerfile
```diff
@@ -14,7 +14,7 @@
 
 # Use the official lightweight Python image.
 # https://hub.docker.com/_/python
-FROM python:3.11-slim
+FROM python:3.10-slim
```
averikitsch, line 17: We want this to be as new as possible to reduce the risk of security and vulnerability issues.

Yuan325:
obsolete: will update retrieval-service python version to 3.11 instead.

PR #107, opened 2023-11-28
chore: add fallback gha workflow for required tests

averikitsch:


kurtisvg (changes requested):
Can this be merged into the existing test file?

averikitsch:
> Can this be merged into the existing test file?

No. We can either remove the docs filter on the other file or have this file. This is only required since the tests are required checks.

PR #126, opened 2023-12-08
feat: add mock for retrieval service test

Yuan325:
Update retrieval service testing to use mock datastore.


retrieval_service/app/app_test.py
```diff
@@ -12,97 +12,329 @@
...
-            kind="postgres",
-            user=DB_USER,
-            password=DB_PASS,
-            database=DB_NAME,
-            host=IPv4Address(DB_HOST),
-        )
-    )
-    app = init_app(cfg)
+    mock_app_config = MagicMock()
+    app = init_app(mock_app_config)
```
kurtisvg, line 211: nit: 
```suggestion
    app = init_app(MagicMock())
```

retrieval_service/app/app_test.py
```diff
@@ -12,97 +12,329 @@
...
+            "city": "BAR",
+            "country": "FOO BAR",
+        },
+        id="iata_only",
+    ),
+]
+
+
+@pytest.mark.parametrize("params, expected", get_airport_params)
+@patch.object(datastore, "create", AsyncMock(return_value=MockDatastore()))
```
kurtisvg, line 254: If it's not too much trouble, I think it's probably cleaner to move from using a single Mock to just using MagicMock and declaring what you want the return values to be using the mock directly in the test. It ties the return value closer to the test -- not only is this easier to update, but means you can have two different tests rely on two different mock values.

kurtisvg (changes requested):

kurtisvg (approved):

PR #137, opened 2023-12-13
feat: Create individual user client session

duwenxin99:


langchain_tools_demo/main.py
```diff
@@ -24,10 +25,20 @@
...
 from starlette.middleware.sessions import SessionMiddleware
 
-from agent import init_agent
-from tools import session
+from agent import init_agent, session
 
-app = FastAPI()
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
```
kurtisvg, line 32: need to close all agents here

langchain_tools_demo/agent.py
```diff
@@ -21,20 +21,23 @@
...
+from tools import convert_date, initialize_tools
+import aiohttp
 
 set_verbose(bool(os.getenv("DEBUG", default=False)))
+session = None
 
 
 # Agent
-def init_agent() -> AgentExecutor:
+async def init_agent() -> AgentExecutor:
```
kurtisvg, line 32: This should probably return AgentExecutor and ClientSession in a new object so we retain references to them both

langchain_tools_demo/agent.py
```diff
@@ -21,20 +21,23 @@
 from langchain.memory import ConversationBufferMemory
 from langchain.prompts.chat import ChatPromptTemplate
 
-from tools import convert_date, tools
+from tools import convert_date, initialize_tools
+import aiohttp
 
 set_verbose(bool(os.getenv("DEBUG", default=False)))
+session = None
```
kurtisvg, line 28: no need to declare a global session here

langchain_tools_demo/agent.py
```diff
@@ -21,20 +21,23 @@
...
-def init_agent() -> AgentExecutor:
+async def init_agent() -> AgentExecutor:
     """Load an agent executor with tools and LLM"""
     print("Initializing agent..")
     llm = VertexAI(max_output_tokens=512)
     memory = ConversationBufferMemory(
         memory_key="chat_history", input_key="input", output_key="output"
     )
-
+    session = aiohttp.ClientSession()
```
kurtisvg, line 39: nit: consider tuning default parameters -- max connections, min connections, etc

langchain_tools_demo/tools.py
```diff
@@ -24,20 +24,13 @@
...
-# create a new client session
-async def get_session():
-    if session is None:
-        client_session = aiohttp.ClientSession()
-    return client_session
 
 
 # Helper functions
-async def get_request(url: str, params: dict) -> aiohttp.ClientResponse:
+async def get_request(
```
kurtisvg, line 30: You may be able to remove this function -- and add the bearer token (if it exists) on session creation

langchain_tools_demo/tools.py
```diff
@@ -94,46 +87,58 @@ class AirportIdInput(BaseModel):
...
-    if response.status != 200:
-        return f"Error trying to find airport: {response}"
+async def generate_get_airport(session: aiohttp.ClientSession):
+    async def get_airport(id: int):
+        response = await get_request(
+            session,
+            f"{BASE_URL}/airports",
+            {"id": id},
+        )
+        if response.status != 200:
```
kurtisvg, line 97: nit: consider using `raise_for_status()` here.

kurtisvg (dismissed):

langchain_tools_demo/tools.py
duwenxin99: removed

langchain_tools_demo/tools.py
duwenxin99: set raise_for_status function on client initializaiton.

langchain_tools_demo/main.py
duwenxin99: close all agents async

langchain_tools_demo/agent.py
duwenxin99: deleted

langchain_tools_demo/agent.py
duwenxin99: created new class ClientAgent

langchain_tools_demo/agent.py
duwenxin99: currently using default max = 100 but could adjust if later found any issues

langchain_tools_demo/agent.py
```diff
@@ -13,28 +13,115 @@
...
 
 set_verbose(bool(os.getenv("DEBUG", default=False)))
+BASE_URL = os.getenv("BASE_URL", default="http://127.0.0.1:8080")
+
+# aiohttp context
+connector = None
+
+
+class ClientAgent:
+    client: aiohttp.ClientSession
```
averikitsch, line 40: We should add a comment on why we are setting up ClientSessions

langchain_tools_demo/agent.py
```diff
@@ -13,28 +13,115 @@
...
+from tools import initialize_tools
 
 set_verbose(bool(os.getenv("DEBUG", default=False)))
+BASE_URL = os.getenv("BASE_URL", default="http://127.0.0.1:8080")
+
+# aiohttp context
+connector = None
+
+
+class ClientAgent:
```
averikitsch, line 39: The ClientAgent name is hard to reason about, could we use something like AgentSession, UserAgent, UserSession, etc?

langchain_tools_demo/agent.py
```diff
@@ -13,28 +13,115 @@
...
+async def handle_error_response(response):
+    if response.status != 200:
+        return f"Error sending {response.method} request to {str(response.url)}): {await response.text()}"
+
+
+async def create_client_session() -> aiohttp.ClientSession:
+    return aiohttp.ClientSession(
+        connector=await get_connector(),
+        headers=get_header(),
+        raise_for_status=handle_error_response,
```
averikitsch, line 111: Let's keep an eye out if the nicer error message worked better than this.

averikitsch (changes requested):

langchain_tools_demo/agent.py
duwenxin99: added comment

langchain_tools_demo/agent.py
duwenxin99: changed to UserAgent

langchain_tools_demo/agent.py
duwenxin99: We could modify the error handler function if we think more details are needed in future usage

averikitsch (approved):

PR #140, opened 2023-12-14
feat: add search airport tools to langchain agent

Yuan325:


langchain_tools_demo/tools.py
```diff
@@ -106,6 +106,37 @@ async def get_airport(id: int):
...
+        {
+            "country": country,
+            "city": city,
+            "name": name,
+        },
+    )
+    if response.status != 200:
+        return f"Error searching airports: {response}"
+
+    num = 2
```
averikitsch, line 127: nit: Do we need to truncate the response for airports? For flights it can list many results, but I don't expect the same issue here. The agent seems to struggle with 5-10+.

averikitsch (approved):

langchain_tools_demo/tools.py
Yuan325: Yes, user only have to provide at least one of (country, city, or name). So if the user only query with one criteria, it's very likely the database return 100s of rows.

PR #141, opened 2023-12-14
feat: add cloudsql integration test

Yuan325:
also update alloydb ci/cd workflow to ignore cloudsql integration test.

This PR does not include adding cloud sql CI/CD cloudbuild workflow. Will add it in a separate PR.

### DIFF
--- a/retrieval_service/alloydb.tests.cloudbuild.yaml
+++ b/retrieval_service/alloydb.tests.cloudbuild.yaml
@@ -74,7 +74,7 @@ steps:
         # Set env var expected by tests
         export DB_USER=$$PGUSER
         export DB_PASS=$$PGPASSWORD
-        python -m pytest -m "not cloudsql"
+        python -m pytest datastore/providers/postgres_test.py
 
   - id: Run database export
     name: python:3.11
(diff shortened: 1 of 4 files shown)


retrieval_service/alloydb.tests.cloudbuild.yaml
```diff
@@ -74,7 +74,7 @@ steps:
         # Set env var expected by tests
         export DB_USER=$$PGUSER
         export DB_PASS=$$PGPASSWORD
-        python -m pytest
+        python -m pytest -m "not cloudsql"
```
averikitsch, line 77: Long term it may be more clear to specify the exact tests like `pytest datastore/providers/postgres_test.py`. We don't have to run the other mocked tests during this integration tests but we will have to set up another config to run the mocked tests.

retrieval_service/datastore/providers/postgres_test.py
```diff
@@ -26,7 +26,7 @@
 from .test_data import query_embedding1, query_embedding2, query_embedding3
 from .utils import get_env_var
 
-pytestmark = pytest.mark.asyncio(scope="module")
+pytestmark = [pytest.mark.asyncio(scope="module"), pytest.mark.postgres]
```
averikitsch, line 29: remove additions if you agree with above comment

retrieval_service/pyproject.toml
```diff
@@ -8,3 +8,9 @@ warn_unused_configs = true
 [[tool.mypy.overrides]]
 module = ["pgvector.asyncpg"]
 ignore_missing_imports = true
+
+[tool.pytest.ini_options]
+markers = [
```
averikitsch, line 13: remove additions if you agree with above comment

averikitsch (approved):

PR #143, opened 2023-12-15
feat: add cloud sql cloudbuild workflow

Yuan325:


retrieval_service/sql-proxy/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,161 @@
...
+        diff --strip-trailing-cr -Z flights_dataset.csv flights_dataset.csv.new || (echo "flight dataset export fail" && exit 1)
+
+  - id: Clean exported files
+    name: python:3.11
+    dir: data
+    entrypoint: /bin/bash
+    args:
+      - "-c"
+      - |
+        rm airport_dataset.csv.new amenity_dataset.csv.new flights_dataset.csv.new
```
averikitsch, line 128: nit: this is not needed. Since it's not stored in Google (like in Cloud Storage or another service), it will be cleaned up automatically after the test run.

retrieval_service/sql-proxy/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,161 @@
...
+        sed -i "/127.0.0.1/d" config.yml
+        sed -i "s/postgres/cloudsql-postgres/g" config.yml
+        sed -i "s/my_database/${_DATABASE_NAME}/g" config.yml
+        sed -i "s/my-user/$$CLOUDSQLUSER/g" config.yml
+        sed -i "s/my-password/$$CLOUDSQLPASSWORD/g" config.yml
+        echo -e '\n  project: "'$PROJECT_ID'"' >> config.yml
+        echo '  region: "'${_CLOUDSQL_REGION}'"' >> config.yml
+        echo '  instance: "'${_CLOUDSQL_INSTANCE}'"' >> config.yml
+        # Run script
+        python run_database_init.py
```
averikitsch, line 84: We will need to start the proxy again in this step

retrieval_service/sql-proxy/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - CLOUDSQLPASSWORD
+    args:
+      - "-c"
+      - |
+        /cloudsql/cloud-sql-proxy --port ${_DATABASE_PORT} ${_INSTANCE_CONNECTION_NAME} & sleep 2;
+        psql -h ${_DATABASE_HOST} -c "CREATE DATABASE ${_DATABASE_NAME};" || echo "Database '${_DATABASE_NAME}' already exists."
+        psql -h ${_DATABASE_HOST} -d ${_DATABASE_NAME} -c 'CREATE EXTENSION vector;' || echo "Extension 'vector' already exists."
+
+  - id: Initialize data
+    name: python:3.11
```
averikitsch, line 64: needs to use the new proxy image

retrieval_service/sql-proxy/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,161 @@
...
+        # Create config
+        cp example-config.yml config.yml
+        sed -i "/127.0.0.1/d" config.yml
+        sed -i "s/postgres/cloudsql-postgres/g" config.yml
+        sed -i "s/my_database/${_DATABASE_NAME}/g" config.yml
+        sed -i "s/my-user/$$CLOUDSQLUSER/g" config.yml
+        sed -i "s/my-password/$$CLOUDSQLPASSWORD/g" config.yml
+        echo -e '\n  project: "'$PROJECT_ID'"' >> config.yml
+        echo '  region: "'${_CLOUDSQL_REGION}'"' >> config.yml
+        echo '  instance: "'${_CLOUDSQL_INSTANCE}'"' >> config.yml
```
averikitsch, line 82: Do we want to add another example config file in order make this easier? We could have rename the current to postgres-config.yml and add another maybe named cloudsql-config.yml

retrieval_service/sql-proxy/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM gcr.io/gcp-runtimes/ubuntu_20_0_4
```
averikitsch, line 1: This Dockerfile needs to contain the code, tests, and dependencies. We will set a placeholder command but  in the Cloud Build we will use a different entrypoint
```suggestion
# Set python runtime
FROM python:3.11-slim

# Copy local code to the container image.
WORKDIR /app
COPY . ./

# Install dependencies.
RUN pip install --no-cache-dir -r requirements.txt -r requirements-test.txt

# Set placeholder command
CMD pytest datastore/providers/cloudsql_postgres_test.py

```

averikitsch: We most likely need this file in the parent directory of retrieval service

retrieval_service/sql-proxy/cloudsql.tests.cloudbuild.yaml
Yuan325: resolving this in another PR. (remove from the alloydb workflow)

Yuan325:
Will add the auto checks to run cloudsql integration test in github after this PR is submitted.

retrieval_service/Dockerfile-cloudsql
```diff
@@ -0,0 +1,37 @@
...
+
+# Set python runtime
+FROM python:3.11-slim
+
+RUN apt-get update \
+  && DEBIAN_FRONTEND=noninteractive \
+  apt-get install --no-install-recommends --assume-yes \
+  postgresql-client
+
+COPY --from=gcr.io/cloud-sql-connectors/cloud-sql-proxy /cloud-sql-proxy /cloudsql/cloud-sql-proxy
```
kurtisvg, line 23: We shouldn't need the proxy for the cloudsql provider right?

kurtisvg (changes requested):

retrieval_service/Dockerfile-cloudsql
```diff
@@ -0,0 +1,37 @@
...
+
+# Copy local code to the container image.
+WORKDIR /app
+
+# Install dependencies.
+COPY ./requirements.txt requirements.txt
+COPY ./requirements-test.txt requirements-test.txt
```
kurtisvg, line 30: nit: I think you can one line this 
```suggestion
COPY ./requirements.txt ./requirements-test.txt ./
```

retrieval_service/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,143 @@
...
+    name: ${_IMAGE_NAME}
+    dir: retrieval_service
+    secretEnv:
+      - PGUSER
+      - PGPASSWORD
+    entrypoint: /bin/bash
+    args:
+      - "-c"
+      - |
+        /cloudsql/cloud-sql-proxy --port ${_DATABASE_PORT} ${_INSTANCE_CONNECTION_NAME} & sleep 2;
```
kurtisvg, line 64: don't need proxy here

retrieval_service/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,143 @@
...
+      - "DB_REGION=${_CLOUDSQL_REGION}"
+      - "DB_INSTANCE=${_CLOUDSQL_INSTANCE}"
+    secretEnv:
+      - PGUSER
+      - PGPASSWORD
+    entrypoint: /bin/bash
+    args:
+      - "-c"
+      - |
+        /cloudsql/cloud-sql-proxy --port ${_DATABASE_PORT} ${_INSTANCE_CONNECTION_NAME} & sleep 2;
```
kurtisvg, line 91: don't need proxy here

retrieval_service/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,143 @@
...
+        python -m pytest datastore/providers/cloudsql_postgres_test.py
+
+  - id: Run database export
+    name: ${_IMAGE_NAME}
+    dir: retrieval_service
+    entrypoint: /bin/bash
+    args:
+      - "-c"
+      - |
+        /cloudsql/cloud-sql-proxy --port ${_DATABASE_PORT} ${_INSTANCE_CONNECTION_NAME} & sleep 2;
```
kurtisvg, line 104: dont need proxy here

kurtisvg (changes requested):

retrieval_service/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+substitutions:
+  _DATABASE_NAME: test_${SHORT_SHA}
+  _DATABASE_USER: postgres
+  _DATABASE_HOST: 127.0.0.1
+  _CLOUDSQL_REGION: "us-central1"
+  _CLOUDSQL_INSTANCE: "my-cloudsql-instance"
+  _INSTANCE_CONNECTION_NAME: ${PROJECT_ID}:${_CLOUDSQL_REGION}:${_CLOUDSQL_INSTANCE}
+  _DATABASE_PORT: "5432"
+  _IMAGE_NAME: gcr.io/${PROJECT_ID}/cloudsql-test-proxy
```
averikitsch, line 113: nit: I apologize for missing this, but GCR will be deprecated this year and we should use Artifact Registry instead. That means we just need to update this image url and create a Docker repo in Artifact Registry

averikitsch (approved):

retrieval_service/cloudsql.tests.cloudbuild.yaml
averikitsch: This can happen now or later

Yuan325: Removing the need to create another Dockerfile instead. :)

retrieval_service/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,84 @@
...
+        # Create config
+        cp example-config-cloudsql.yml config.yml
+        sed -i "s/my_database/${_DATABASE_NAME}/g" config.yml
+        sed -i "s/my-user/$$PGUSER/g" config.yml
+        sed -i "s/my-password/$$PGPASSWORD/g" config.yml
+        sed -i "s/my-project/$PROJECT_ID/g" config.yml
+        sed -i "s/my-region/${_CLOUDSQL_REGION}/g" config.yml
+        sed -i "s/my-instance/${_CLOUDSQL_INSTANCE}/g" config.yml
+
+  - id: Run Alloy DB integration tests
```
averikitsch, line 49: Update database name

retrieval_service/cloudsql.tests.cloudbuild.yaml
```diff
@@ -0,0 +1,84 @@
+# Copyright 2023 Google LLC
```
kurtisvg, line 1: nit: 2024 now

retrieval_service/datastore/providers/cloudsql_postgres_test.py
```diff
@@ -59,15 +58,47 @@ def db_instance() -> str:
...
+    sys_conn: asyncpg.Connection = await connector.connect_async(
+        f"{db_project}:{db_region}:{db_instance}",
+        "asyncpg",
+        user=f"{db_user}",
+        password=f"{db_pass}",
+        db="postgres",
+    )
+    await sys_conn.execute(f'DROP DATABASE IF EXISTS "{db_name}";')
+    await sys_conn.execute(f'CREATE DATABASE "{db_name}";')
+    await sys_conn.close()
```
kurtisvg, line 78: nit: should probably use:
```
async with sys_conn:
    await sys_conn.execute()
```
so it closes if there is an error

retrieval_service/datastore/providers/cloudsql_postgres_test.py
```diff
@@ -59,15 +58,47 @@ def db_instance() -> str:
...
 async def ds(
+    create_db: AsyncGenerator[str, None],
     db_user: str,
     db_pass: str,
-    db_name: str,
     db_project: str,
     db_region: str,
     db_instance: str,
 ) -> AsyncGenerator[datastore.Client, None]:
+    db_name = await create_db.__anext__()
```
kurtisvg, line 101: Do we really need to do __anext__ like that?

kurtisvg (approved):

retrieval_service/datastore/providers/cloudsql_postgres_test.py
Yuan325: Yes, this is the only way that I can get it to work since create_db is returning an async generator (async with yield)

PR #144, opened 2023-12-18
feat: update amenity embedding to include amenity name

Yuan325:


averikitsch:
Just to check did you use run_generate_embeddings.py? I just want to make sure we have the correct embedding model version.

averikitsch:
Can you update the demo database once merged?

averikitsch (approved):

PR #148, opened 2023-12-21
fix: update search airport tool

averikitsch:


langchain_tools_demo/agent.py
```diff
@@ -70,20 +70,6 @@ def get_id_token(url: str) -> str:
         )
 
 
-def convert_date(date_string: str) -> str:
```
Yuan325, line 73: Are we assuming that the agent will be able to do the date conversion automatically?

Yuan325 (approved):

langchain_tools_demo/agent.py
averikitsch: I was unable to get the agent to use this as a tool

duwenxin99 (approved):

langchain_tools_demo/agent.py
Yuan325: Ahhh, got it~

PR #151, opened 2023-12-27
feat: update amenity_dataset with new hour columns

Yuan325:
Update amenity_dataset to include 14 new column for start_hour and end_hour of each day.
Update postgres provider, cloudsql provider and tests.


retrieval_service/app/app_test.py
```diff
@@ -250,6 +255,20 @@ def test_search_airports_with_bad_params(m_datastore, app, params):
             terminal="FOO BAR",
             category="FEE",
             hour="BAZ",
+            sunday_start_hour=get_time_mock(),
```
kurtisvg, line 258: I don't understand why this is referencing a fixture as a function, or why we need a function here as opposed to inlining it?

retrieval_service/models/models.py
```diff
@@ -38,9 +38,43 @@ class Amenity(BaseModel):
     terminal: str
     category: str
     hour: str
+    sunday_start_hour: Optional[datetime.time]
```
kurtisvg, line 41: Does this need to be Optional?

retrieval_service/models/models.py
```diff
@@ -38,9 +38,43 @@ class Amenity(BaseModel):
...
+        "thursday_start_hour",
+        "thursday_end_hour",
+        "friday_start_hour",
+        "friday_end_hour",
+        "saturday_start_hour",
+        "saturday_end_hour",
+        mode="before",
+    )
+    def replace_none(cls, v):
+        return v or None
```
kurtisvg, line 76: Is this still neccesary with an optional field?

kurtisvg (changes requested):

retrieval_service/models/models.py
Yuan325: Yes, if the store is not operating on that day, this value will be None.

retrieval_service/models/models.py
Yuan325: Yes, within the csv dataset, these `None` value are empty. For example: 
Sunday 9-10pm, 09:00:00, 22:00:00,,,,, (those `None` value is converted as just empty).
Hence, when running validation, if without replacing them with actual `None`, it will throw an error saying that the type is different.

retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -140,7 +154,7 @@ async def initialize_data(
             # Insert all the data
             await conn.execute(
                 text(
-                    """INSERT INTO amenities VALUES (:id, :name, :description, :location, :terminal, :category, :hour, :content, :embedding)"""
+                    """INSERT INTO amenities VALUES (:id, :name, :description, :location, :terminal, :category, :hour, :sunday_start_hour, :sunday_end_hour, :monday_start_hour, :monday_end_hour, :tuesday_start_hour, :tuesday_end_hour, :wednesday_start_hour, :wednesday_end_hour, :thursday_start_hour, :thursday_end_hour, :friday_start_hour, :friday_end_hour, :saturday_start_hour, :saturday_end_hour, :content, :embedding)"""
```
kurtisvg, line 157: nit: this is a long line, maybe try to break it out into a few different lines

kurtisvg: (same with somes othes)

retrieval_service/models/models.py
kurtisvg: I think that's okay, but we probably need to update the retrieval service to say "closed" as opposed to None in the string it returns.

kurtisvg (approved):

retrieval_service/models/models.py
Yuan325: added a bug for this

PR #152, opened 2023-12-27
feat: Create reset button to clear session cookies

duwenxin99:


langchain_tools_demo/main.py
```diff
@@ -95,6 +93,19 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
...
+async def reset(request: Request):
+    """Reset agent"""
+    global user_agents
+    tasks = []
+    for user_agent in user_agents.values():
+        tasks.append(asyncio.create_task(user_agent.client.close()))
+        del user_agent.agent
+    asyncio.gather(*tasks)
+    user_agents.clear()
+    request.session.clear()
```
kurtisvg, line 106: This logic looks like it will delete and close all the agents. I think what we want instead is to just wipe the information for that user/session. This probably means:

1. grabbing the uuid from the session
2. use the uuid to grab the user agents
3. close the agent
3. remove uuid/agent from `user_agents`
4. remove the uuid and messages from the session

kurtisvg (changes requested):

langchain_tools_demo/main.py
duwenxin99: changed to only deleting the current user agent.

langchain_tools_demo/main.py
```diff
@@ -95,6 +93,20 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
         raise HTTPException(status_code=500, detail=f"Error invoking agent: {err}")
 
 
+@app.post("/reset")
+async def reset(request: Request):
+    """Reset agent"""
+    global user_agents
+    uuid = request.session["uuid"]
+    if uuid in user_agents.keys():
```
kurtisvg, line 101: if uuid isn't in user_agents, should we throw an error?

langchain_tools_demo/main.py
```diff
@@ -95,6 +93,20 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
...
+        user_agent = user_agents[uuid]
+        await user_agent.client.close()
+        del user_agent.agent
+        del user_agents[request.session["uuid"]]
+    request.session.clear()
+    print(user_agents)
```
kurtisvg, line 107: Is this for debugging? Should it stay in the current implementation?

kurtisvg (changes requested):

langchain_tools_demo/main.py
```diff
@@ -95,6 +93,20 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
...
 
+@app.post("/reset")
+async def reset(request: Request):
+    """Reset agent"""
+    global user_agents
+    uuid = request.session["uuid"]
+    if uuid in user_agents.keys():
+        user_agent = user_agents[uuid]
+        await user_agent.client.close()
+        del user_agent.agent
```
kurtisvg, line 104: why we are deleting just the agent here?

kurtisvg (changes requested):

langchain_tools_demo/main.py
duwenxin99: I was thinking it would be helpful for the user to learn which agents are currently active. But I could remove this line if it is not necessary.

langchain_tools_demo/main.py
duwenxin99: added deleting agent client

langchain_tools_demo/main.py
duwenxin99: That way you would get an error when you click the reset button first before start chatting with the agent (the agent hasn't been initialized yet). This may not be desired.

kurtisvg: I thought the agent was initialized when we load the index ("/") -- doesn't that mean the agent should always been initalized? Are we redirecting them to the index after we call this endpoint?

kurtisvg (changes requested):

langchain_tools_demo/main.py
duwenxin99: Currently the agent is initialized in the "/chat" handler on line 83. We could move this part to the index (I think it would be more intuitive)

kurtisvg: Ah GitHub shrank the sections for me so I missed the function declaration. 

I agree -- let's move it to index.

duwenxin99: moved to index and raised exception in "/reset" if agent not found.

langchain_tools_demo/main.py
```diff
@@ -70,18 +74,10 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
...
     # Add user message to chat history
     request.session["messages"] += [{"role": "user", "content": prompt}]
-    # Agent setup
-    if request.session["uuid"] in user_agents:
-        user_agent = user_agents[request.session["uuid"]]
-    else:
-        user_agent = await init_agent()
-        user_agents[request.session["uuid"]] = user_agent
+
+    user_agent = user_agents[(request.session["uuid"])]
```
kurtisvg, line 80: nit: 
```suggestion
    user_agent = user_agents[request.session["uuid"]]
```

langchain_tools_demo/main.py
```diff
@@ -95,6 +91,19 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
...
+@app.post("/reset")
+async def reset(request: Request):
+    """Reset agent"""
+    global user_agents
+    uuid = request.session["uuid"]
+    if uuid in user_agents.keys():
+        await user_agents[uuid].client.close()
+        del user_agents[uuid]
+    else:
+        raise HTTPException(status_code=500, detail=f"Current agent not found")
```
kurtisvg, line 103: nit:

you should flip this and do:

if uuid not in X:
    raise Exception
    
# happy path here

kurtisvg (approved):

langchain_tools_demo/main.py
duwenxin99: fixed

langchain_tools_demo/main.py
duwenxin99: fixed

PR #156, opened 2024-01-02
chores: move db_init and db_export as a datastore function

Yuan325:


retrieval_service/datastore/datastore.py
```diff
@@ -44,6 +47,76 @@ def kind(cls):
...
+    async def export_dataset(
+        self,
+        airports,
+        amenities,
+        flights,
+        airports_new_path,
+        amenities_new_path,
+        flights_new_path,
+    ) -> None:
+        np.set_printoptions(linewidth=100000)
```
kurtisvg, line 78: Do we still need this line?

kurtisvg (approved):

PR #188, opened 2024-01-25
feat: vertex ai function calling llm

Yuan325:
move langchain to a new demo folders and add function calling.


langchain_tools_demo/orchestrator/vertexai_function_calling/function_calling_orchestrator.py
```diff
@@ -0,0 +1,158 @@
...
+            model_response = self.chat.send_message(prompt)
+        except Exception as err:
+            raise HTTPException(status_code=500, detail=f"Error invoking agent: {err}")
+        return model_response
+
+    async def request_function(self, function_call):
+        url = function_request(function_call["name"])
+        params = function_call["args"]
+        print(f"function url is {url}")
+        print(f"params is {params}")
```
kurtisvg, line 93: should these be logs or removed?

langchain_tools_demo/orchestrator/vertexai_function_calling/function_calling_orchestrator.py
```diff
@@ -0,0 +1,158 @@
...
+            url=f"{BASE_URL}/{url}",
+            params=params,
+            headers=get_headers(self.client),
+        )
+        response = await response.json()
+        return response
+
+
+class FunctionCallingOrchestrator(BaseOrchestrator):
+    _user_sessions: Dict[str, UserChatModel] = {}
```
kurtisvg, line 104: should this be a class field? Or an object field?

kurtisvg (changes requested):
Would be good to add some tests if we can

langchain_tools_demo/orchestrator/vertexai_function_calling/function_calling_orchestrator.py
Yuan325: hmm, for the current design seems like we are only generating a single Orchestrator class per app initiation. If we will allow switch between multiple Orchestrator per app in the future then I can change it back to class field. I'll move this definition to the init function! Thankyou!

kurtisvg (approved):

PR #196, opened 2024-01-29
fix: missing Client ID

averikitsch:


duwenxin99 (approved):

langchain_tools_demo/main.py
```diff
@@ -66,13 +66,12 @@ async def index(request: Request):
...
     # Agent setup
     agent = await get_agent(request.session)
-    print(request.session["history"])
     return templates.TemplateResponse(
         "index.html",
         {
             "request": request,
             "messages": request.session["history"],
-            "client_id": request.session.get("client_id"),
+            "client_id": os.getenv("CLIENT_ID"),
```
kurtisvg, line 74: Can we move this to the file level? I don't think we want to do a syscall on every index call 
```suggestion
CLIENT_ID = os.getenv("CLIENT_ID")
# ...
            "client_id": CLIENT_ID,
```

kurtisvg (approved):

PR #197, opened 2024-01-29
doc: Update instruction to set CLIENT_ID env var

duwenxin99:


docs/run_langchain_demo.md
```diff
@@ -32,6 +32,11 @@
     pip install -r requirements.txt
     ```
 
+1. [Create a Client ID](https://support.google.com/cloud/answer/6158849) for your app and save it as an environment variable:
```
kurtisvg, line 35: This is optional right? Can we say something like "if you want to take advantage of the User authentication features, you'll need to set up Ouath2:"

kurtisvg (changes requested):

docs/run_langchain_demo.md
duwenxin99: Made it optional

kurtisvg (approved):

averikitsch (approved):

Issue #190, opened 2024-01-27
I am getting HTTP error 500 in my app when I run it. It started yesterday. 


gotochkin:
Here is from the debug log:

Finished chain.
Action:
```
{
  "action": "Search Airport",
  "action_input": {
    "country": "Canada",
    "city": "Ottawa",
    "name": null
  }
}
```INFO:     127.0.0.1:46522 - "POST /chat HTTP/1.1" 500 Internal Server Error

Can we troubleshoot it? Looks like some of the latest updates have broken it.

**Comment by kurtisvg** [comment]
I found the issue by running on a GCE engine account using the default service account. It then get's the following error:

```
  File "/home/gleb/genai-databases-retrieval-app/langchain_tools_demo/tools.py", line 46, in get_headers
    headers["Authorization"] = f"Bearer {get_id_token()}"
                                         ^^^^^^^^^^^^^^
  File "/home/gleb/genai-databases-retrieval-app/langchain_tools_demo/tools.py", line 38, in get_id_token
    return CREDENTIALS.id_token
           ^^^^^^^^^^^^^^^^^^^^
```

The TL;DR is that the `CREDENTIALS.id_token` when the token it retrieved with the VM service account. Looking at the docs for ID_Token, it looks like compute engine has a [slightly different flow](https://cloud.google.com/docs/authentication/get-id-token#python). We probably need to implement that flow:
```python
request = google.auth.transport.requests.Request()
    # Set the target audience.
    # Setting "use_metadata_identity_endpoint" to "True" will make the request use the default application
    # credentials. Optionally, you can also specify a specific service account to use by mentioning
    # the service_account_email.
    credentials = compute_engine.IDTokenCredentials(
        request=request, target_audience=url, use_metadata_identity_endpoint=True
    )

    # Get the ID token.
    # Once you've obtained the ID token, use it to make an authenticated call
    # to the target audience.
    credentials.refresh(request)
    # print(credentials.token)
    print("Generated ID token.")
```

kurtisvg:
(I think we can check if credentials exists in ID_Token, and if it doesn't we can use this other call).

PR #198, opened 2024-01-30
fix: Add ID Token credential flow for GCE

duwenxin99:
Fixes: https://github.com/GoogleCloudPlatform/genai-databases-retrieval-app/issues/190

### DIFF
--- a/langchain_tools_demo/tools.py
+++ b/langchain_tools_demo/tools.py
@@ -40,10 +40,15 @@ def get_id_token():
     if hasattr(CREDENTIALS, "id_token"):
         return CREDENTIALS.id_token
     else:
-        # Use Compute Engine default credential 
-        CREDENTIALS = compute_engine.IDTokenCredentials(request=Request(), target_audience=BASE_URL, use_metadata_identity_endpoint=True)
+        # Use Compute Engine default credential
+        CREDENTIALS = compute_engine.IDTokenCredentials(
+            request=Request(),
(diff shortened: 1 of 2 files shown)


langchain_tools_demo/tools.py
```diff
@@ -36,7 +37,17 @@ def get_id_token():
...
+        return CREDENTIALS.id_token
+    else:
+        # Use Compute Engine default credential
+        CREDENTIALS = compute_engine.IDTokenCredentials(
+            request=Request(),
+            target_audience=BASE_URL,
+            use_metadata_identity_endpoint=True,
+        )
+        CREDENTIALS.refresh(Request())
+        return CREDENTIALS.token
```
kurtisvg, line 50: I think there is a small bug on refresh:

Consider the following flow:

1. get_id_token is called 
2. CREDENTIALS are set to google.auth.default()
3. CREDENTIALS doesn't have id_token
4. CREDENTIALS is replaced w/ IDToken credentials 
5. CREDENTIALS.token is returned
6.  -- some time passes --
7. get_id_token is called 
8. CREDENTIALS doesn't have id_token
9. CREDENTIALS is replaced w/ IDToken credentials

kurtisvg (changes requested):

langchain_tools_demo/tools.py
duwenxin99: modified

kurtisvg (approved):

Issue #206, opened 2024-02-07
The run_database_init.py returns error


gotochkin:
Here is the error: 

Traceback (most recent call last):
  File "/home/gleb/genai-databases-retrieval-app/retrieval_service/run_database_init.py", line 39, in <module>
    asyncio.run(main())
  File "/home/gleb/.pyenv/versions/3.11.6/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/home/gleb/.pyenv/versions/3.11.6/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gleb/.pyenv/versions/3.11.6/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/gleb/genai-databases-retrieval-app/retrieval_service/run_database_init.py", line 27, in main
    cfg = parse_config("config.yml")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gleb/genai-databases-retrieval-app/retrieval_service/app/app.py", line 41, in parse_config
    return AppConfig(**config)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/gleb/.pyenv/versions/3.11.6/lib/python3.11/site-packages/pydantic/main.py", line 164, in __init__
    __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
pydantic_core._pydantic_core.ValidationError: 1 validation error for AppConfig
clientId
  Field required [type=missing, input_value={'host': '0.0.0.0', 'data...': 'ShittyPassword01#'}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.5/v/missing

It seems clientId parameter is missing in the config.yml template. It was stated as an optional but the parse_config cannot really skip it if it is present in the class AppConfig

Can we please fix it? I am getting complains from users trying to execute codelab and qwiklab. 
Thanks,
Gleb

duwenxin99:
Fixed.

PR #207, opened 2024-02-07
fix: Make `clientId `optional in config.yml

duwenxin99:
fix issue: https://github.com/GoogleCloudPlatform/genai-databases-retrieval-app/issues/206

### DIFF
--- a/retrieval_service/example-config.yml
+++ b/retrieval_service/example-config.yml
@@ -8,3 +8,4 @@ datastore:
   database: "my_database"
   user: "my-user"
   password: "my-password"
+clientId: "my-clientId"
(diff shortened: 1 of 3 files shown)


retrieval_service/example-config.yml
```diff
@@ -7,4 +7,5 @@ datastore:
   # port: 5432
   database: "my_database"
   user: "my-user"
-  password: "my-password"
\ No newline at end of file
+  password: "my-password"
+clientId: "my-clientId"
```
kurtisvg, line 11:
```suggestion
  password: "my-password"
  # clientId: "my-client-id"
```

kurtisvg (approved):

PR #225, opened 2024-02-12
feat: update app in langchain demo

Yuan325:
Update existing app to have less global state within main.py
Add run_app.py
Add example-config.yml

user will still be able to run app via `python main.py`. Current implementation won't be affected.


kurtisvg (changes requested):

langchain_tools_demo/main.py
```diff
@@ -153,6 +149,17 @@ async def reset(request: Request):
     request.session.clear()
 
 
+def init_app() -> FastAPI:
+    # FastAPI setup
+    app = FastAPI(lifespan=lifespan)
+    app.include_router(routes)
+    app.mount("/static", StaticFiles(directory="static"), name="static")
+    # TODO: set secret_key for production
+    app.add_middleware(SessionMiddleware, secret_key="SECRET_KEY")
```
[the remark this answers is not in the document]
Yuan325: SessionMiddleware must be installed to access request.session. Do you meant just removing the secret_key for now?

langchain_tools_demo/main.py
```diff
@@ -153,6 +149,17 @@ async def reset(request: Request):
     request.session.clear()
 
 
+def init_app() -> FastAPI:
```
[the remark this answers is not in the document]
Yuan325: Updated~ and added AppConfig.

langchain_tools_demo/main.py
```diff
@@ -32,10 +33,31 @@
...
 from agent import init_agent, user_agents
 
+BASE_HISTORY: list[BaseMessage] = [
+    AIMessage(content="I am an SFO Airport Assistant, ready to assist you.")
+]
+routes = APIRouter()
+
+
+class AppConfig(BaseModel):
+    host: IPv4Address | IPv6Address = IPv4Address("0.0.0.0")
```
kurtisvg, line 49: nit: should we default to 127.0.0.1?

langchain_tools_demo/example-config.yml
```diff
@@ -0,0 +1,5 @@
+host: 0.0.0.0
+port: 8081
```
kurtisvg, line 2: should this be $PORT?

langchain_tools_demo/main.py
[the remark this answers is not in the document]
kurtisvg: Ah I guess we should actually set a secret_key somewhere then. Maybe an env var?

langchain_tools_demo/main.py
```diff
@@ -153,6 +165,20 @@ async def reset(request: Request):
...
+    # TODO: set secret_key for production
+    app.add_middleware(SessionMiddleware, secret_key="SECRET_KEY")
+    return app
+
+
 if __name__ == "__main__":
     PORT = int(os.getenv("PORT", default=8081))
+    cfg = parse_config("./config.yml")
+    app = init_app(cfg)
     uvicorn.run(app, host="0.0.0.0", port=PORT)
```
kurtisvg, line 184: I think this is ignoring the port set in the app

langchain_tools_demo/main.py
```diff
@@ -50,34 +72,24 @@ async def lifespan(app: FastAPI):
...
-
-
-@app.route("/", methods=["GET", "POST"])
+@routes.get("/")
+@routes.post("/")
 async def index(request: Request):
     """Render the default template."""
     # Agent setup
     agent = await get_agent(request.session, user_id_token=None)
+    templates = Jinja2Templates(directory="templates")
```
kurtisvg, line 81: nit: I think it's okay for this to be declared globally. It's essentially a static property

langchain_tools_demo/run_app.py
```diff
@@ -0,0 +1,37 @@
...
+
+import asyncio
+import os
+
+import uvicorn
+
+from main import init_app, parse_config
+
+
+async def main():
```
kurtisvg, line 24: this should probably be mostly the same function as what's here: https://github.com/GoogleCloudPlatform/genai-databases-retrieval-app/pull/225/files#diff-1bce6b474df8443122d1e8581ae869c7222de09f43c0b3037ef573815daeeb74R180

kurtisvg (changes requested):

langchain_tools_demo/main.py
Yuan325: Updating this. I'm keeping this `if __name__` for now, but eventually we will remove it during the last step. Wanted to make sure that this changes does not cause any change to existing instructions yet.

langchain_tools_demo/main.py
Yuan325: Cloud build will fail with 127.0.0.1. Will keep it as 0.0.0.0

kurtisvg (approved):

PR #226, opened 2024-02-13
feat: add orchestration interface

Yuan325:
Add orchestration interface and move langchain orchestration.


langchain_tools_demo/orchestration/orchestration.py
```diff
@@ -0,0 +1,82 @@
...
+    @abstractmethod
+    async def create(cls, history: List[Any]) -> "Orchestration":
+        pass
+
+    async def invoke(self, prompt: str):
+        raise NotImplementedError("Subclass should implement this!")
+
+    @staticmethod
+    async def get_connector():
+        global connector
```
kurtisvg, line 50: Can we make this a class property instead?

langchain_tools_demo/orchestration/orchestration.py
```diff
@@ -0,0 +1,82 @@
...
+# aiohttp context
+connector = None
+
+
+class classproperty:
+    def __init__(self, func):
+        self.fget = func
+
+    def __get__(self, instance, owner):
+        return self.fget(owner)
```
kurtisvg, line 29: I think we can use `@classmethod @property`: https://stackoverflow.com/questions/128573/using-property-on-classmethods/64738850#64738850

langchain_tools_demo/orchestration/orchestration.py
```diff
@@ -0,0 +1,82 @@
...
+    async def get_connector():
+        global connector
+        if connector is None:
+            connector = aiohttp.TCPConnector(limit=100)
+        return connector
+
+    @staticmethod
+    async def handle_error_response(response):
+        if response.status != 200:
+            return f"Error sending {response.method} request to {str(response.url)}): {await response.text()}"
```
kurtisvg, line 58: If we want a static method we can just put it as a utility somewhere -- we don't need to nest on the object

langchain_tools_demo/orchestration/orchestration.py
```diff
@@ -0,0 +1,82 @@
...
+
+class classproperty:
+    def __init__(self, func):
+        self.fget = func
+
+    def __get__(self, instance, owner):
+        return self.fget(owner)
+
+
+class Orchestration(ABC):
```
kurtisvg, line 32: nit:
```suggestion
class BaseOrchestrator(ABC):
```

langchain_tools_demo/orchestration/langchain_tools/agent.py
```diff
@@ -14,101 +14,101 @@
...
 
-# aiohttp context
-connector = None
-
 CLOUD_RUN_AUTHORIZATION_TOKEN = None
 
 
-# Class for setting up a dedicated llm agent for each individual user
-class UserAgent:
+class Orchestration(orchestration.Orchestration):
```
kurtisvg, line 39:
```suggestion
class LangChainToolsOrchestrator(BaseOrchestrator):
```

langchain_tools_demo/orchestration/langchain_tools/agent.py
```diff
@@ -14,101 +14,101 @@
...
     client: aiohttp.ClientSession
     agent: AgentExecutor
 
-    def __init__(self, client, agent) -> None:
+    @orchestration.classproperty
```
kurtisvg, line 43:
```suggestion
    @classmethod
    @property
```

langchain_tools_demo/main.py
```diff
@@ -118,34 +114,38 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
...
         )
         return markdown(response["output"])
     except Exception as err:
         raise HTTPException(status_code=500, detail=f"Error invoking agent: {err}")
 
 
-async def get_agent(session: dict[str, Any], user_id_token: Optional[str]):
-    global user_agents
+async def get_agent(
+    session: dict[str, Any], user_id_token: Optional[str], orchestration: Optional[str]
```
kurtisvg, line 133: Why let orchestration be optional?

langchain_tools_demo/orchestration/orchestration.py
```diff
@@ -0,0 +1,82 @@
...
+            connector_owner=False,
+            headers={},
+            raise_for_status=True,
+        )
+
+    def close(self):
+        raise NotImplementedError("Subclass should implement this!")
+
+
+ais: Dict[str, Orchestration] = {}
```
kurtisvg, line 73: nit: this is global state -- can we move it into a class somewhere?

langchain_tools_demo/orchestration/orchestration.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+async def create(orchestration: str, history: List[Any]) -> Orchestration:
+    for cls in Orchestration.__subclasses__():
+        s = f"{orchestration} == {cls.kind}"
+        if orchestration == cls.kind:
+            return await cls.create(history)  # type: ignore
+    raise TypeError(f"No orchestration of kind {orchestration}")
+    return
```
kurtisvg, line 82:
```suggestion
```

langchain_tools_demo/orchestration/langchain_tools/agent.py
```diff
@@ -14,101 +14,101 @@
...
-    template = "\n\n".join([PREFIX, tool_strings, format_instructions, SUFFIX, today])
-    human_message_template = "{input}\n\n{agent_scratchpad}"
-    prompt = ChatPromptTemplate.from_messages(
-        [("system", template), ("human", human_message_template)]
-    )
-    agent.agent.llm_chain.prompt = prompt  # type: ignore
-
-    return UserAgent(client, agent)
+    @classmethod
+    async def create(cls, base_history: list[Any]) -> "Orchestration":
```
kurtisvg, line 52: We need to separate the "Orchestration" from the agents inside of it. The "Orchestrator" layer should be able to handle everything, including getting agents

kurtisvg (changes requested):

langchain_tools_demo/orchestration/orchestration.py
Yuan325: Seems like this feature was removed in Python 3.13. :(

langchain_tools_demo/orchestration/orchestration.py
Yuan325: Removed this function instead. Seems like we no longer use this.

langchain_tools_demo/main.py
Yuan325: We call for the get_agent function in 3 different endpoints. and within the 3 function calls, technically only one of them will be creating the agent. Hence, similar to user_id_token, I'm thinking of passing the orchestration variable when needed.

langchain_tools_demo/orchestration/orchestration.py
Yuan325: Since this is used to keep track of the llm for each user, the only way to remove it from global is to set it as app.state. Let me know if theres any other better way :).

langchain_tools_demo/main.py
```diff
@@ -49,19 +36,17 @@ async def lifespan(app: FastAPI):
...
 
 @routes.get("/")
 @routes.post("/")
 async def index(request: Request):
     """Render the default template."""
     # Agent setup
-    agent = await get_agent(request.session, user_id_token=None)
+    orchestrator = request.app.state.orchestrator
+    await orchestrator.user_session_create(request.session)
+    templates = Jinja2Templates(directory="templates")
```
kurtisvg, line 49: nit: this should be global state since it's static. We don't want to reload on every call to index.

langchain_tools_demo/main.py
```diff
@@ -139,19 +103,23 @@ async def reset(request: Request):
         raise HTTPException(status_code=400, detail=f"No session to reset.")
 
     uuid = request.session["uuid"]
-    global user_agents
-    if uuid not in user_agents.keys():
+    orchestrator = request.app.state.orchestrator
+    if not orchestrator.user_session_exist(uuid):
         raise HTTPException(status_code=500, detail=f"Current agent not found")
```
kurtisvg, line 108:
```suggestion
        raise HTTPException(status_code=500, detail=f"Current user session not found")
```

langchain_tools_demo/main.py
```diff
@@ -161,9 +129,10 @@ def init_app(client_id: Optional[str], secret_key: Optional[str]) -> FastAPI:
 if __name__ == "__main__":
     PORT = int(os.getenv("PORT", default=8081))
     HOST = os.getenv("HOST", default="0.0.0.0")
+    ORCHESTRATOR = os.getenv("ORCHESTRATOR")
```
kurtisvg, line 132: nit: s/ORCHESTRATOR/ORCHESTRATION_TYPE/

langchain_tools_demo/orchestrator/langchain_tools/langchain_tools_orchestrator.py
```diff
@@ -0,0 +1,234 @@
...
+        )
+        return prompt
+
+    def parse_messages(self, datas: List[Any]) -> List[BaseMessage]:
+        messages: List[BaseMessage] = []
+        for data in datas:
+            if data["type"] == "human":
+                messages.append(HumanMessage(content=data["data"]["content"]))
+            if data["type"] == "ai":
+                messages.append(AIMessage(content=data["data"]["content"]))
```
kurtisvg, line 169: It's almost always a bad idea to just ignore bad data. Log or return an error instead so their is some feedback someone did something wrong. 
```suggestion
            if data["type"] == "human":
                messages.append(HumanMessage(content=data["data"]["content"]))
            elif data["type"] == "ai":
                messages.append(AIMessage(content=data["data"]["content"]))
            else:
                raise Exception()
```

PR #238, opened 2024-02-27
chore: Update prompt for list flight

Yuan325:


llm_demo/orchestrator/langchain_tools/tools.py
```diff
@@ -352,6 +352,7 @@ async def initialize_tools(client: aiohttp.ClientSession):
             description="""
                         Use this tool to list a user's flight tickets.
                         Takes no input and returns a list of current user's flight tickets.
+                        Input is always empty JSON blob.
```
kurtisvg, line 355: would it be better to provide an example here?

kurtisvg (approved):

PR #249, opened 2024-02-27
feat: update to include user name in chat

Yuan325:


llm_demo/app.py
```diff
@@ -66,11 +68,20 @@ async def login_google(
...
     if user_id_token is None:
         raise HTTPException(status_code=401, detail="No user credentials found")
+
+    user_name = get_user_name(str(user_id_token), request.app.state.client_id)
     # create new request session
     orchestrator = request.app.state.orchestration_type
     orchestrator.set_user_session_header(request.session["uuid"], str(user_id_token))
     print("Logged in to Google.")
 
+    request.session["history"][0] = {
```
kurtisvg, line 78: This is replacing the previous history (which is misleading). Should we either:
a. only do this if there is one message in the chat
or
b. append to the end to indicate the user is signed on?

kurtisvg (changes requested):

llm_demo/app.py
kurtisvg: should we also move it into the orchestration layer -- isn't that where the history is being handled now?

Yuan325: Nope, the history is being handled by the app request.

kurtisvg (approved):

PR #258, opened 2024-03-07
feat: implement signin signout

Yuan325:


llm_demo/orchestrator/orchestrator.py
```diff
@@ -61,6 +61,15 @@ def set_user_session_header(self, uuid: str, user_id_token: str):
         user_session = self.get_user_session(uuid)
         user_session.client.headers["User-Id-Token"] = f"Bearer {user_id_token}"
 
+    def get_user_id_token(self, uuid: str) -> Optional[str]:
+        user_session = self.get_user_session(uuid)
+        if user_session.client and "User-Id-Token" in user_session.client.headers:
+            token = user_session.client.headers["User-Id-Token"]
+            parts = str(token).split(" ")
+            if len(parts) == 2:
```
kurtisvg, line 69: nit: should their be more visibility here if something goes wrong?

Maybe logging or throwing an error would be better then silently failing

kurtisvg (approved):

PR #264, opened 2024-03-12
feat: Add flight prices.

tabbyl21:
In this commit we add prices to the flights data set. Prices are based on fligh duration, and centered around $200 with a mean of the flight duration and standard deviation of prices based on flight duration.


Yuan325 (approved):
LGTM, thank you! I will add make the code changes to load the flight price column into our databases :).

PR #265, opened 2024-03-13
feat: add cymbal air passenger policy

Yuan325:


llm_demo/orchestrator/cymbalair_policy.py
```diff
@@ -0,0 +1,33 @@
+POLICY = """Policy that are listed below is unchangeable. You will not answer any questions outside of the following policy.
+Cymbal Air: Passenger Policy
+----------------------------
+
+1.  Ticket Purchase and Changes
```
averikitsch, line 5: nit: using markdown headers might be better than using an ordered list

llm_demo/orchestrator/vertexai_function_calling/function_calling_orchestrator.py
```diff
@@ -93,7 +94,7 @@ async def invoke(self, input_prompt: str) -> Dict[str, Any]:
     def get_prompt(self) -> str:
         formatter = "%A, %m/%d/%Y, %H:%M:%S"
         now = datetime.now(timezone("US/Pacific")).strftime("%A, %m/%d/%Y, %H:%M:%S")
-        prompt = f"{PREFIX}. Today's date and current time is {now}."
+        prompt = f"{PREFIX} {POLICY} Today's date and current time is {now}."
```
averikitsch, line 97: Do we need to add newlines?

kurtisvg (approved):

PR #266, opened 2024-03-13
chore: fix firestore import test fail

Yuan325:


averikitsch (approved):
We still need a workflow to run these tests, correct?

PR #267, opened 2024-03-14
(feat) Add Seats

mahyareb:
- Add a seats table to the schema
- Add id to flights table since seats requires it
- Load data from GCS Buckets instead of CSV because files are now too large
- preload both tickets and seats when building dataset
- Add a seat to the booking during the flight booking transaction
- Add initial tool to check if a specific flight has a seat of user's preference available.

Screenshot 2024-03-14 at 4 18 37 PM
Screenshot 2024-03-14 at 4 20 00 PM
Screenshot 2024-03-14 at 4 19 10 PM


retrieval_service/datastore/providers/postgres.py
```diff
@@ -425,6 +500,8 @@ async def insert_ticket(
         arrival_airport: str,
         departure_time: str,
         arrival_time: str,
+        seat_row: int | None = None,
```
Yuan325, line 503: Can we also add test (in postgres_test.py) for insert_tickets and list_tickets since we now have a dataset for it.

docs/datastore/alloydb.md
```diff
@@ -236,7 +236,7 @@ Private IP.
       password: "my-alloydb-pass"
     ```
 
-1. Populate data into database:
+1. Populate data into database. Note: This make take up to 5 minutes.
```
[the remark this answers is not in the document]
mahyareb: Good idea, done!

retrieval_service/datastore/providers/postgres.py
mahyareb: Great idea, done!

llm_demo/orchestrator/langchain_tools/tools.py
```diff
@@ -223,6 +223,38 @@ async def list_tickets():
     return list_tickets
 
 
+def generate_list_seats(client: aiohttp.ClientSession):
```
Yuan325, line 226: Might wanna add a `class SeatInput(BaseModel)` (can refer to the other input class at the top).

llm_demo/orchestrator/langchain_tools/tools.py
```diff
@@ -355,4 +387,50 @@ async def initialize_tools(client: aiohttp.ClientSession):
...
+                            "airline": "OO",
+                            "flight_number": "6307",
+                            "departure_airport": "SFO",
+                            "departure_time": "2024-10-28 20:13:00",
+                            "seat_row": 25,
+                            "seat_letter": "B",
+                            "seat_class": null,
+                            "seat_type": null,
+                        }}
+                        """,
```
Yuan325, line 434: After adding the `SeatInput` class, add `args_schema=SeatInput` here. (can refer to the other tool above. List ticket doesn't have the schema since it doesn't require any input.

llm_demo/orchestrator/langchain_tools/tools.py
Yuan325: I'm also wondering if we want to add `seat_row` and `seat_letter` as optional input in the `insert_ticket` tools above. (I'm assuming we will allow user to insert ticket without picking a seat? Or do they have to pick a seat?

If change, also need to update the examples for this tool below.

retrieval_service/models/models.py
```diff
@@ -105,3 +106,15 @@ class Ticket(BaseModel):
     arrival_airport: str
     departure_time: datetime.datetime
     arrival_time: datetime.datetime
+    seat_row: int
```
Yuan325, line 109: Is this required or optional?

retrieval_service/app/routes.py
```diff
@@ -148,6 +148,32 @@ async def search_flights(
     return flights
 
 
+@routes.get("/seats/search")
+async def search_seats(
```
Yuan325, line 152: Sorry I missed this earlier, but we'll also have to update app_test.py for this endpoint.

retrieval_service/datastore/providers/postgres.py
```diff
@@ -215,14 +218,76 @@ async def initialize_data(
...
+                ],
+            )
+
+            # If the table already exists, drop it to avoid conflicts
+            await conn.execute("DROP TABLE IF EXISTS seats CASCADE")
+            # Create a new table
+            await conn.execute(
+                """
+                CREATE TABLE seats(
+                  flight_id INTEGER,
```
Yuan325, line 255: Nit: Does this flight_id corresponds to the `id` column in flights table? I'm wondering will it be better to update the `id` column to `flight_id` for flights database. (not sure how much this will help the llm agent to understand things better).

mahyareb: I think it's normal and best practice to have "id" in a table x end up being "x_id" in other tables. Same with ticket_id below.

llm_demo/orchestrator/langchain_tools/tools.py
mahyareb: Awesome, sounds good! Done!

llm_demo/orchestrator/langchain_tools/tools.py
mahyareb: Great idea, done! And adding seat_row and seat_letter to insert_ticket as well in the next commit just shortly.

retrieval_service/datastore/providers/postgres.py
Yuan325: Got it~ sounds good!

retrieval_service/app/routes.py
mahyareb: Done!

Yuan325 (approved):
Final nit, can you also update the path in`retrieval_service/run_generate_embeddings.py` for the amenities dataset since the dataset is moved to gcs? Other than that, LGTM :)

PR #268, opened 2024-03-14
chore: update prompt and amenity dataset

Yuan325:
- update list flight description to make date as required. (or it will throw an exception)
- update one of the amenities name

PR #272, opened 2024-03-19
feat: add time filter

Yuan325:


llm_demo/orchestrator/langchain_tools/tools.py
```diff
@@ -155,13 +155,24 @@ async def list_flights(
 
 class QueryInput(BaseModel):
     query: str = Field(description="Search query")
+    filter_time: str = Field(description="Time for filtering amenities operating hours")
```
kurtisvg, line 158: nit: Consider renaming "open_time" and "open_day". It's the time/day we want them to be open. 
```suggestion
    filter_time: str = Field(description="Time for filtering amenities by operating hours.")
```

retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -330,17 +330,34 @@ async def get_amenity(self, id: int) -> Optional[models.Amenity]:
...
+        query_embedding: list[float],
+        similarity_threshold: float,
+        top_k: int,
+        filter_time: Optional[str],
+        filter_day: Optional[str],
     ) -> list[models.Amenity]:
+
+        filter_time_datetime = None
+        filter_query = "WHERE "
+        if filter_time and filter_day:
```
kurtisvg, line 343: What happens if this is one but not the other? will it just be ignored?

retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -330,17 +330,34 @@ async def get_amenity(self, id: int) -> Optional[models.Amenity]:
...
+            start_hour = filter_day + "_start_hour"
+            end_hour = filter_day + "_end_hour"
```
kurtisvg, line 345: Should we validate the possible options that filter_day should be?

retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -330,17 +330,34 @@ async def get_amenity(self, id: int) -> Optional[models.Amenity]:
...
+            filter_time_datetime = datetime.strptime(filter_time, "%H:%M:%S").time()
+            filter_query += f""" {start_hour} <= CAST(:filter_time AS TIME)
+                      AND {end_hour} > CAST(:filter_time AS TIME)
```
kurtisvg, line 348: Any reason we aren't parsing this on the python side?

kurtisvg (changes requested):

llm_demo/orchestrator/langchain_tools/langchain_tools_orchestrator.py
```diff
@@ -266,6 +275,14 @@ def close_clients(self):
 }}}}
 ```"""
 
+MULTITURN_PROMPT = """Assistant will judge if multiple action need to be taken before returning final response to human.
```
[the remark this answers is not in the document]
Yuan325: Yeah, that was in another PR that's merged earlier. (https://github.com/GoogleCloudPlatform/genai-databases-retrieval-app/pull/268)

retrieval_service/datastore/providers/cloudsql_postgres.py
Yuan325: It's checked in routes.py. Will throw an error if one is given but not the other.

retrieval_service/datastore/providers/cloudsql_postgres.py
Yuan325: Added the filter in route.py. Seems like theres where most endpoint input validations were made.

kurtisvg (approved):

retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -330,17 +330,34 @@ async def get_amenity(self, id: int) -> Optional[models.Amenity]:
...
+        query_embedding: list[float],
+        similarity_threshold: float,
+        top_k: int,
+        open_time: Optional[str],
+        open_day: Optional[str],
     ) -> list[models.Amenity]:
+
+        open_time_datetime = None
+        filter_query = "WHERE "
+        if open_time and open_day:
```
kurtisvg, line 343: What happens if they specified one and not the other?

kurtisvg (approved):

retrieval_service/datastore/providers/cloudsql_postgres.py
Yuan325: at the endpoint level it will throw an error. Within the providers, it will just not use the filter.

PR #303, opened 2024-03-29
Add Tracing

mahyareb:
Add Tracing from Retrieval service and NL2Query service.

https://screencast.googleplex.com/cast/NTc5NzM4MTgwNjIyNzQ1NnxlMDNlODMwNy01Yg

### DIFF
--- a/llm_demo/orchestrator/langchain_tools/langchain_tools_orchestrator.py
+++ b/llm_demo/orchestrator/langchain_tools/langchain_tools_orchestrator.py
@@ -98,7 +98,9 @@ class UserAgent:
         self.agent.tools = tools
 
     async def insert_ticket(self, params: str, tool_trace: ToolTrace):
-        return await insert_ticket(self.client, params, tool_trace)
+        result = await insert_ticket(self.client, params, tool_trace)
+        trace = tool_trace.flush()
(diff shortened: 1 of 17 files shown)


llm_demo/orchestrator/langchain_tools/langchain_tools_orchestrator.py
```diff
@@ -148,10 +152,12 @@ async def user_session_create(self, session: dict[str, Any]):
         user_email = ""
         if "user_info" in session:
             user_email = session["user_info"].get("user_email", "")
-        tools = await initialize_tools(client, user_email)
+        tool_trace = ToolTrace()
+        tools = await initialize_tools(client, user_email, tool_trace)
```
Yuan325, line 156: Need to update at the `set_user_email` function below too. It re-initializes the tools with user_email when user logs in.

Yuan325 (commented):
Thank you!! This is great!

llm_demo/orchestrator/langchain_tools/langchain_tools_orchestrator.py
mahyareb: Ah yeah, I missed this in a merge conflict. Good catch! :) Fixed!

PR #306, opened 2024-03-29
fix(ci): use script over args in cloudbuild syntax

kurtisvg:


mahyareb (approved):

Yuan325 (approved):
I'll add the same changes to other cloud build files~ Thanks!

PR #307, opened 2024-03-29
fix(ci): use script over args in all cloudbuild

Yuan325:


retrieval_service/cloudsql.tests.cloudbuild.yaml
```diff
@@ -16,27 +16,16 @@ steps:
...
     dir: retrieval_service
     secretEnv:
-      - PGUSER
-      - PGPASSWORD
-    entrypoint: /bin/bash
-    args:
-      - "-c"
-      - |
+      - DB_USER
+      - DB_PASS
```
kurtisvg, line 26: You will need to update these in the sed commands as well

kurtisvg (approved):

PR #329, opened 2024-04-04
fix: sign out when user token invalid

Yuan325:
when the user token is invalid, signout to clear the session instead of just clearning session['user_info'].


llm_demo/orchestrator/.orchestrator.py.swp
averikitsch, line 1: Is this needed?

llm_demo/orchestrator/orchestrator.py
```diff
@@ -66,20 +66,22 @@ def set_user_session_header(self, uuid: str, user_id_token: str):
...
     def get_user_id_token(self, uuid: str) -> Optional[str]:
-        user_session = self.get_user_session(uuid)
-        if user_session.client and "User-Id-Token" in user_session.client.headers:
-            token = user_session.client.headers["User-Id-Token"]
-            parts = str(token).split(" ")
-            if len(parts) != 2 or parts[0] != "Bearer":
-                raise Exception("Invalid ID token")
-            return parts[1]
+        if self.user_session_exist(uuid):
+            user_session = self.get_user_session(uuid)
```
averikitsch, line 70: nit: this could be combined with the line above

llm_demo/orchestrator/orchestrator.py
```diff
@@ -66,20 +66,22 @@ def set_user_session_header(self, uuid: str, user_id_token: str):
...
         return None
 
     async def user_session_signout(self, uuid: str):
         """Sign out from user session. Clear and restart session."""
         user_session = self.get_user_session(uuid)
-        await user_session.close()
-        del user_session
+        if user_session:
+            await user_session.close()
+            del user_session
```
averikitsch, line 84: Doesn't this need to be deleted from the dictionary like `del self._user_sessions[user_session]`?

averikitsch (approved):

llm_demo/orchestrator/.orchestrator.py.swp
Yuan325: whoops, thanks!

PR #336, opened 2024-04-08
Fix issue with gemini temperature

mahyareb:


AndrewBrook-Google (commented):
Let's use ISO 8601 date format consistently, please.  YYYY-MM-DD

AndrewBrook-Google (approved):
Let's use ISO 8601 date format consistently, please.  YYYY-MM-DD

LGTM now!

PR #357, opened 2024-04-17
update and add tests

Yuan325:


retrieval_service/datastore/providers/alloydb_test.py
```diff
@@ -0,0 +1,730 @@
...
+    # Database does not exist, create it.
+    sys_conn: asyncpg.Connection = await connector.connect(
+        f"projects/{db_project}/locations/{db_region}/clusters/{db_cluster}/instances/{db_instance}",
+        "asyncpg",
+        user=f"{db_user}",
+        password=f"{db_pass}",
+        db="postgres",
+        ip_type="PUBLIC",
+    )
+    print("didn't even get here")
```
kurtisvg, line 91:
```suggestion
```

retrieval_service/datastore/providers/alloydb_test.py
```diff
@@ -0,0 +1,730 @@
...
+    )
+
+    diff_airports = compare(
+        load_csv(open(airports_ds_path), "id"), load_csv(open(airports_new_path), "id")
+    )
+    assert diff_airports["added"] == []
+    assert diff_airports["removed"] == []
+    assert diff_airports["changed"] == []
+    assert diff_airports["columns_added"] == []
+    assert diff_airports["columns_removed"] == []
```
kurtisvg, line 182: nit: given this is repeated 3 times, should it be in a function instead?

kurtisvg (approved):

PR #367, opened 2024-05-01
feat: add tracing to main

Yuan325:
Add tracing to main. This PR add tracing for postgres.py provider and langchain orchestration.

### DIFF
--- a/llm_demo/app.py
+++ b/llm_demo/app.py
@@ -154,12 +154,17 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
     response = await orchestrator.user_session_invoke(request.session["uuid"], prompt)
     output = response.get("output")
     confirmation = response.get("confirmation")
+    trace = response.get("trace")
     # Return assistant response
     if confirmation:
-        return json.dumps({"type": "confirmation", "content": confirmation})
+        return json.dumps(
+            {"type": "confirmation", "content": confirmation, "trace": trace}
+        )
     else:
         request.session["history"].append({"type": "ai", "data": {"content": output}})
-        return json.dumps({"type": "message", "content": markdown(output)})
+        return json.dumps(
+            {"type": "message", "content": markdown(output), "trace": trace}
+        )
 
 
 @routes.post("/book/flight", response_class=PlainTextResponse)
@@ -180,7 +185,7 @@ async def book_flight(request: Request, params: str = Body(embed=True)):
     request.session["history"].append(
         {"type": "ai", "data": {"content": "I have booked your ticket."}}
     )
-    return response
+    return json.dumps(response)
 
 
 @routes.post("/book/flight/decline", response_class=PlainTextResponse)
@@ -198,7 +203,7 @@ async def decline_flight(request: Request):
 
 
 @routes.post("/reset")
-def reset(request: Request):
+async def reset(request: Request):
     """Reset user session"""
 
     if "uuid" not in request.session:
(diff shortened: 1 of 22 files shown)


llm_demo/app.py
```diff
@@ -180,7 +185,7 @@ async def book_flight(request: Request, params: str = Body(embed=True)):
     request.session["history"].append(
         {"type": "ai", "data": {"content": "I have booked your ticket."}}
     )
-    return response
+    return json.dumps(response)
```
kurtisvg, line 188: should the other endpoints return json as well?

llm_demo/orchestrator/helpers/tool_trace.py
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List
+
+
+class ToolTrace:
```
kurtisvg, line 18: Can we add some comments? What is this class, and why is it helping?

llm_demo/orchestrator/langchain_tools/langchain_tools_orchestrator.py
```diff
@@ -160,19 +171,37 @@ async def user_session_invoke(self, uuid: str, prompt: str) -> dict[str, Any]:
...
+        # clear session history
         del session["history"]
         base_history = self.get_base_history(session)
         session["history"] = [base_history]
         history = self.parse_messages(session["history"])
         user_session.reset_memory(history)
 
+        # clear session traces
+        del self._user_traces[uuid]
+        tool_trace = ToolTrace()
```
kurtisvg, line 189: do we need to hang on to ToolTrace if it's used by tools?

llm_demo/orchestrator/helpers/tool_trace.py
```diff
@@ -0,0 +1,30 @@
...
+    curr_log: List[str] = []
+
+    def __init__(self):
+        self.curr_log = []
+
+    def add_message(self, message: str):
+        self.curr_log.append(message)
+
+    def flush(self):
```
kurtisvg, line 27: what does flush do?

llm_demo/orchestrator/langchain_tools/tools.py
```diff
@@ -83,15 +85,18 @@ async def search_airports(country: str, city: str, name: str):
 
         num = 2
         response_json = await response.json()
-        if len(response_json) < 1:
+        if response_json.get("trace"):
+            tool_trace.add_message(response_json.get("trace"))
```
kurtisvg, line 89:
```suggestion
       trace = response_json.get("trace")
        if trace:
            tool_trace.add_message(trace)
```

kurtisvg: Do we need to validate trace is in the right format somewhere?

kurtisvg (changes requested):

llm_demo/app.py
Yuan325: Updated the return from retrieval service.

llm_demo/orchestrator/langchain_tools/langchain_tools_orchestrator.py
Yuan325: ToolTrace is connected to each user session (to prevent it from sending the wrong tooltrace to another user. (it is similar to _user_sessions. There is another way to do this by initializing it and hanging on to it within UserAgent. What do you think about that?

llm_demo/orchestrator/langchain_tools/tools.py
Yuan325: "trace" are build in retrieval service/app/routes.py (src code is in: retrieval_service/helpers/helpers.py) before sending it back to the UI.

PR #368, opened 2024-05-01
chore: add tracing to vertex ai function calling

Yuan325:


llm_demo/orchestrator/vertexai_function_calling/functions.py
```diff
@@ -201,8 +205,10 @@ async def insert_ticket(client: aiohttp.ClientSession, params: str):
         },
         headers=get_headers(client),
     )
-    response = await response.json()
-    return response
+    response_json = await response.json()
+    if response_json.get("trace"):
+        tool_trace.add_message(response_json.get("trace"))
```
kurtisvg, line 210: nit: better to only call 'get' once

kurtisvg (approved):

PR #375, opened 2024-05-02
Select stable model https://ai.google.dev/gemini-api/docs/models/gemi…

mahyareb:
Revert to a stable version of gemini and remove time for predictive outcomes.


AndrewBrook-Google:
Note - we are having issues with the demo.  Suspect Gemini version update may have introduced regressions.

PR #394, opened 2024-05-22
chore: add versioning

Yuan325:


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,18 @@
+on:
```
kurtisvg, line 1: nit: please add licensing header

kurtisvg (approved):

PR #411, opened 2024-06-13
chore: add golden dataset for eval

Yuan325:
Adding golden datasets that will be used in llm system evaluation.

The golden dataset is separated into multiple types of queries:
- queries that uses a specific tool
- airline related queries (no tool calling - answer is within prompt)
- assistant related question (no tool calling - answer is within prompt)
- out of context questions (no tool calling)
- multitool selections (agent selecting multiple tool before returning a final answer to the user)

### DIFF
--- a/llm_demo/evaluation/eval_golden.py
+++ b/llm_demo/evaluation/eval_golden.py
@@ -1,0 +1,316 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+goldens = [
+    {
+        "Search Airport Tool": [
+            {
+                "query": "What is the airport located in san francisco?",
+                "tool_calls": [
+                    {
+                        "name": "Search Airport",
+                        "arguments": {"city": "san francisco"},
+                    },
+                ],
+                "reset": True,
+            },
+            {
+                "query": "Tell me more about denver international airport?",
+                "tool_calls": [
+                    {
+                        "name": "Search Airport",
+                        "arguments": {
+                            "city": "denver",
+                            "name": "denver international airport",
+                        },
+                    },
+                ],
+                "reset": True,
+            },
+        ]
+    },
+    {
+        "Search Flights By Flight Number Tool": [
+            {
+                "query": "What is the departure gate for flight CY 922?",
+                "tool_calls": [
+                    {
+                        "name": "Search Flights By Flight Number",
+                        "arguments": {
+                            "airline": "CY",
+                            "flight_number": "922",
+                        },
+                    },
+                ],
+                "reset": True,
+            },
+            {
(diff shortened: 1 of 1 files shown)


jackwotherspoon:
Just putting this here as a friendly suggestion, can take it or leave it 😄 

Long-term open source maintainability can improve drastically by PRs having at a minimum a one or two sentence description (especially for `feat` PRs but even for all types). Just going over what you are adding and why.

The benefits:
-  Helps your reviewer quickly understand the PR
- Helps future maintainers follow-along with PRs
- Allows you to more easily search for old PRs in the future
- Encourages open source external contributions as external users better understand the features and history

I recently have found this extremely beneficial in Cloud SQL repos ([example](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/pull/1093))
 
Github also [recommends this themselves](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/best-practices-for-pull-requests#provide-context-and-guidance)

llm_demo/evaluation/eval_golden.py
```diff
@@ -0,0 +1,286 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+goldens = [
+    {
+        "Search Airport Tool": [
+            {
+                "query": "What is the airport located in san francisco?",
```
kurtisvg, line 19: nit: Should we use proper capitalization here?

llm_demo/evaluation/eval_golden.py
```diff
@@ -0,0 +1,286 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+goldens = [
+    {
+        "Search Airport Tool": [
```
kurtisvg, line 17: Curious to the structure -- why is this list -> object {key: object{} }?

Wouldn't it be better to just use an object at the higher level with key -> object?

kurtisvg (changes requested):

kurtisvg:
+1 to Jack's comment. We could do a better job on PR descriptions.

llm_demo/evaluation/eval_golden.py
Yuan325: Each evaluation data object contains of "query, tool call, output etc.". 

And since I am trying to categorize them into different categories (e.g. we'll provide 2 test data for 'Search Airport Tool' so that we can test out 2 different queries that trigger that tool. If we want to add more queries in the future we can just add it to the category's list.), that's why we have list of evaluation data based on the category (a.k.a. key).

llm_demo/evaluation/eval_golden.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+class EvalData(BaseModel):
+    category: Optional[str] = Field(default=None, description="evaluation category")
+    query: Optional[str] = Field(default=None, description="user query")
+    instruction: Optional[str] = Field(
+        default=None, description="instruction to llm system"
+    )
+    content: Optional[str] = Field(default=None)
+    tool_calls: Optional[List[Dict[str, Any]]] = Field(default=None)
```
kurtisvg, line 27: is "List[Dict[str, Any]]" the best option here? Or would ie be better to introduce another model like "ToolCall" or "ToolRecord" and have "List[ToolCall]"?

kurtisvg (approved):

PR #414, opened 2024-06-14
chore: get prediction for eval dataset

Yuan325:
Add the function to get prediction for each of the queries from golden_dataset. Prediction is used as comparison to retrieve metrics.

Usage example:

```
from evaluation import run_llm_for_eval, goldens

# set up orchestration, session, set uuid
eval_list = await run_llm_for_eval(goldens, orchestration, session, session_id)
```

### DIFF
--- a/llm_demo/evaluation/eval_golden.py
+++ b/llm_demo/evaluation/eval_golden.py
@@ -32,7 +32,7 @@ class EvalData(BaseModel):
     )
     content: Optional[str] = Field(default=None)
     tool_calls: Optional[List[ToolCall]] = Field(default=None)
-    context: Optional[str] = Field(
+    context: Optional[List[Any]] = Field(
         default=None, description="context given to llm in order to answer user query"
     )
     output: Optional[str] = Field(default=None)
--- a/llm_demo/evaluation/eval_golden.py
+++ b/llm_demo/evaluation/eval_golden.py
@@ -32,7 +32,7 @@ class EvalData(BaseModel):
     )
     content: Optional[str] = Field(default=None)
     tool_calls: Optional[List[ToolCall]] = Field(default=None)
-    context: Optional[List[Any]] = Field(
+    context: Optional[List[Dict[str, Any] | List[Dict[str, Any]]]] = Field(
         default=None, description="context given to llm in order to answer user query"
     )
     output: Optional[str] = Field(default=None)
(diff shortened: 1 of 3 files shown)


llm_demo/evaluation/evaluation.py
```diff
@@ -0,0 +1,84 @@
...
+from typing import Any, Dict, List, Optional
+
+from pydantic import BaseModel, Field
+
+from orchestrator import BaseOrchestrator
+
+from .eval_golden import goldens
+
+
+class EvalData(BaseModel):
```
kurtisvg, line 25: It looks like we are using pydantic to convert from JSON to a Python object -- any reason to just not declare as a python object?

llm_demo/evaluation/evaluation.py
```diff
@@ -0,0 +1,84 @@
...
+                cases = golden[key]
+                for case in cases:
+                    data = EvalData(**case)
+                    data.category = key
+                    eval_datas.append(data)
+        return eval_datas
+
+    async def run_llm_for_eval(
+        self, eval_datas: List[EvalData], orc: BaseOrchestrator, session: Dict, uid: str
+    ) -> List[EvalData]:
```
kurtisvg, line 61: Why do we need a class for this?

llm_demo/evaluation/evaluation.py
```diff
@@ -0,0 +1,84 @@
...
+            for key in golden:
+                cases = golden[key]
+                for case in cases:
+                    data = EvalData(**case)
+                    data.category = key
+                    eval_datas.append(data)
+        return eval_datas
+
+    async def run_llm_for_eval(
+        self, eval_datas: List[EvalData], orc: BaseOrchestrator, session: Dict, uid: str
```
kurtisvg, line 60: super-nit:
```suggestion
        self, eval_list: List[EvalData], orc: BaseOrchestrator, session: Dict, uid: str
```

kurtisvg (changes requested):

llm_demo/evaluation/evaluation.py
Yuan325: Will update this in the eval-dataset PR.

kurtisvg (approved):

llm_demo/evaluation/eval_golden.py
```diff
@@ -32,7 +32,7 @@ class EvalData(BaseModel):
     )
     content: Optional[str] = Field(default=None)
     tool_calls: Optional[List[ToolCall]] = Field(default=None)
-    context: Optional[str] = Field(
+    context: Optional[List[Any]] = Field(
```
kurtisvg, line 35: Consider something more strict than `Any`

kurtisvg (approved):

PR #417, opened 2024-06-18
ci: Update cicd for cloudsql

Yuan325:


retrieval_service/example-config-cloudsql.yml
```diff
@@ -1,7 +1,7 @@
 host: 0.0.0.0
 datastore:
   # Example for Cloud SQL
-  kind: "cloudsql-postgres"
+  kind: "cloudsql-dialect"
```
kurtisvg, line 4: Why is this dialect and not a specific engine type? They should still be different datastores right?

retrieval_service/datastore/providers/alloydb.py
```diff
@@ -46,7 +46,7 @@ class Client(datastore.Client[Config]):
 
     @datastore.classproperty
     def kind(cls):
-        return "alloydb-postgres"
+        return POSTGRES_IDENTIFIER
```
kurtisvg, line 49: nit: Should probably be "ALLOYDB_PG_IDENTIFIER"

retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -45,7 +45,7 @@ class Client(datastore.Client[Config]):
 
     @datastore.classproperty
     def kind(cls):
-        return "cloudsql-postgres"
+        return POSTGRES_IDENTIFIER
```
kurtisvg, line 48: nit: CLOUD_SQL_PG_IDENTIFIER

retrieval_service/datastore/providers/firestore.py
```diff
@@ -36,7 +38,7 @@ class Client(datastore.Client[Config]):
 
     @datastore.classproperty
     def kind(cls):
-        return "firestore"
+        return DB_IDENTIFIER
```
kurtisvg, line 41:
```suggestion
        return FIRESTORE_IDENTIFIER
```

kurtisvg (changes requested):

retrieval_service/example-config-cloudsql.yml
Yuan325: Update this to be `cloudsql-engine`. The other configs is the same across cloudsql-postgres and cloudsql-mysql so I was thinking we can reduce the duplication of files by using the same ones. This is used during cloudbuild workflow to setup config.yml.

retrieval_service/datastore/providers/cloudsql_postgres_test.py
```diff
@@ -40,12 +40,12 @@
 
 @pytest.fixture(scope="module")
 def db_user() -> str:
-    return get_env_var("DB_USER", "name of a postgres user")
+    return get_env_var("DB_USER_PG", "name of a postgres user")
```
kurtisvg, line 43: do you still want to change this?

kurtisvg (approved):

PR #419, opened 2024-06-18
chore: add retrieval phase evaluation

Yuan325:
Add the function to run evaluation for retrieval phase (evaluating the llm ability to pick and utilize tools that are given).

Usage example:

```
eval_results = evaluate_retrieval_phase(eval_datas)
```

### DIFF
--- a/llm_demo/evaluation/evaluation.py
+++ b/llm_demo/evaluation/evaluation.py
@@ -62,25 +62,13 @@ async def run_llm_for_eval(
     return eval_list
 
 
-def evaluate_task(
-    eval_dataset: "pd.DataFrame", metrics: List[str], experiment_name: str
-) -> evaluation_base.EvalResult:
-    """
-    Run VertexAI Evaluation Task.
-    """
-    eval_task = EvalTask(
-        dataset=eval_dataset,
-        metrics=metrics,
-        experiment=experiment_name,
-    )
-
-    eval_result = eval_task.evaluate()
-    return eval_result
-
-
 def evaluate_retrieval_phase(eval_datas: List[EvalData]) -> evaluation_base.EvalResult:
+    """
+    Run evaluation for the ability of a model to select the right tool and arguments (retrieval phase).
+    """
     RETRIEVAL_EXPERIMENT_NAME = "retrieval-phase-eval"
     metrics = ["tool_call_quality"]
+    # Prepare evaluation task input
     responses = []
     references = []
     for e in eval_datas:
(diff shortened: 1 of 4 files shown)


llm_demo/evaluation/evaluation.py
```diff
@@ -82,3 +86,40 @@ async def run_llm_for_eval(
...
                 orc.user_session_reset(session, uid)
         return eval_datas
+
+    def evaluate_task(
+        self, eval_dataset: "pd.DataFrame", metrics: List[str], experiment_name: str
+    ) -> evaluation_base.EvalResult:
+        """
+        Run VertexAI Evaluation Task.
+        """
+        nest_asyncio.apply()
```
kurtisvg, line 96: did we figure out a way to not use this? I'd prefer not to put this dependency given it's archived

llm_demo/evaluation/evaluation.py
```diff
@@ -82,3 +86,40 @@ async def run_llm_for_eval(
...
+                    {"content": e.content, "tool_calls": e.prediction_tool_calls}
+                )
+            )
+        eval_dataset = pd.DataFrame(
+            {
+                "response": responses,
+                "reference": references,
+            }
+        )
+        return self.evaluate_task(eval_dataset, metrics, RETRIEVAL_EXPERIMENT_NAME)
```
kurtisvg, line 125: this return isn't type annotated -- shouldn't that throw an error?

kurtisvg (changes requested):

llm_demo/evaluation/evaluation.py
Yuan325: This will be updated in Vertex SDK v1.58.0. I'll resume this PR then.

llm_demo/evaluation/evaluation.py
```diff
@@ -53,3 +60,51 @@ async def run_llm_for_eval(
...
+    Run VertexAI Evaluation Task.
+    """
+    eval_task = EvalTask(
+        dataset=eval_dataset,
+        metrics=metrics,
+        experiment=experiment_name,
+    )
+
+    eval_result = eval_task.evaluate()
+    return eval_result
```
kurtisvg, line 78: Is this really worth a function? 

Isn't it the same as just calling:
```    
 eval_task = EvalTask(
      dataset=eval_dataset,
      metrics=metrics,
      experiment=experiment_name,
 ).execute()
```

kurtisvg (approved):

PR #421, opened 2024-06-20
feat: Add Firestore vector embedding support

ferdeleong:
Enable amenities and policies search with the new added Firebase vector k-nearest neighbor in version 2.16.0.  

To support search functionality, it is required to cast amenities and policies field embedding to Vector type and create single-field vector indexing for both collections.

Finally k-nearest neighbor query with DOT_PRODUCT is applied to find the closest semantic similarity.


duwenxin99 (approved):

retrieval_service/requirements.txt
```diff
@@ -1,7 +1,7 @@
 asyncpg==0.29.0
 fastapi==0.109.2
 google-auth==2.32.0
-google-cloud-firestore==2.14.0
+google-cloud-firestore @ git+https://github.com/googleapis/python-firestore.git@2de16209409c9d9ba41d3444400e6a39ee1b2936
```
averikitsch, line 4: The library is now released https://github.com/googleapis/python-firestore/releases. This can be updated

retrieval_service/datastore/providers/firestore.py
```diff
@@ -76,6 +79,68 @@ async def delete_collections(collection_list: list[AsyncCollectionReference]):
             [airports_ref, amenities_ref, flights_ref, policies_ref]
         )
 
+        async def delete_indexes(index_list):
```
averikitsch, line 82: nit: we could probably have this as a private function so it is not nested in this function

retrieval_service/datastore/providers/firestore.py
```diff
@@ -76,6 +79,68 @@ async def delete_collections(collection_list: list[AsyncCollectionReference]):
...
+
+        # Decode and format output
+        indexes = stdout.decode().strip().split("\n")
+
+        # Check if the indexes already exist; if so, delete indexes
+        # Extract collection and index-id from file path
+        # Assign the index-id to the corresponding collection
+        if indexes == [""]:
+            pass
+        else:
```
averikitsch, line 135:
```suggestion
        if indexes != [""]:
```

retrieval_service/datastore/providers/firestore.py
```diff
@@ -76,6 +79,68 @@ async def delete_collections(collection_list: list[AsyncCollectionReference]):
...
+            collections = {"amenities": "", "policies": ""}
+            for line in indexes:
+                collection, index_id = line.split("/")[-3], line.split("/")[-1]
+                collections[collection] = index_id
+            amenities_ref = collections["amenities"]
+            policies_ref = collections["policies"]
+            await delete_indexes([amenities_ref, policies_ref])
```
averikitsch, line 142:
```suggestion
            collections = []
            for line in indexes:
                collection, index_id = line.split("/")[-3], line.split("/")[-1]
                collections.append(index_id)
            await delete_indexes(collections)
```

retrieval_service/datastore/providers/firestore.py
```diff
@@ -105,8 +170,80 @@ async def delete_collections(collection_list: list[AsyncCollectionReference]):
                         "terminal": amenity.terminal,
                         "category": amenity.category,
                         "hour": amenity.hour,
+                        # Firebase does not support datetime.time type
+                        "sunday_start_hour": (
+                            str(amenity.sunday_start_hour)
```
averikitsch, line 175: Do we want to use Timestamp (https://cloud.google.com/python/docs/reference/firestore/latest/google.cloud.firestore_v1.types.Value)?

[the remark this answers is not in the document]
averikitsch: Do we need the if else or can it be simplified using or?

retrieval_service/datastore/providers/firestore.py
```diff
@@ -105,8 +170,80 @@ async def delete_collections(collection_list: list[AsyncCollectionReference]):
...
+                        ),
+                        "saturday_end_hour": (
+                            str(amenity.saturday_end_hour)
+                            if amenity.saturday_end_hour
+                            else None
+                        ),
                         "content": amenity.content,
-                        "embedding": amenity.embedding,
+                        # Vector type does not support None value
+                        "embedding": Vector(tuple(amenity.embedding or [])),
```
averikitsch, line 246: Does this need to be wrapped in a tuple?

retrieval_service/datastore/providers/firestore.py
```diff
@@ -142,12 +283,49 @@ async def delete_collections(collection_list: list[AsyncCollectionReference]):
...
-                        "embedding": policy.embedding,
+                        # Vector type does not accept None value
+                        "embedding": Vector(policy.embedding or []),
                     }
                 )
             )
         await asyncio.gather(*create_policies_tasks)
 
+        # Initialize single-field vector indexes
+        create_amenities_vector_index = [
```
averikitsch, line 294: nit: we could refactor this so we don't have 2 similar lists (1 for amenities and policies). It might be nice to have these outside of this code block

retrieval_service/datastore/providers/firestore.py
```diff
@@ -326,8 +533,23 @@ async def list_tickets(
 
     async def policies_search(
         self, query_embedding: list[float], similarity_threshold: float, top_k: int
-    ) -> list[str]:
-        raise NotImplementedError("Semantic search not yet supported in Firestore.")
+    ) -> list[Any]:
+        collection = AsyncQuery(self.__client.collection("policies"))
+        query_vector = Vector(query_embedding)
+        distance_measure = DistanceMeasure.DOT_PRODUCT
```
averikitsch, line 539: nit: do we want this as a global variable?

retrieval_service/datastore/providers/firestore.py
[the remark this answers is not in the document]
ferdeleong: `or` doesn't work in this case. If `amenity.sunday_syart_hour ` is none then you end up with str(None) or none.  This will cast None into a string.

retrieval_service/requirements.txt
ferdeleong: That release has vector similarity search but not async vector similarity search. Releases had been stopped since there is a recent breach  https://github.com/googleapis/python-firestore/pull/933

retrieval_service/datastore/providers/firestore.py
ferdeleong: I think explicitly stating which collection I'm passing and deleting might be more readable than just sending all the unidentified list of index-ids. Thoughts?

retrieval_service/datastore/providers/firestore.py
ferdeleong: I think Timestamp always includes Date and Time, and these fields have only time (datetime.time). Thoughts?

retrieval_service/datastore/providers/firestore.py
ferdeleong: delete collections is also nested, should I change both?

retrieval_service/requirements.txt
averikitsch: This is now released https://pypi.org/project/google-cloud-firestore/2.17.0/

retrieval_service/datastore/providers/firestore.py
averikitsch: yes, that will make this code more readable

retrieval_service/datastore/providers/firestore.py
averikitsch: I see. We can keep this then for the clarity. Do we need to change anything if "amenities" and "policies" doesn't exist or the name changes and do we not want to delete all indexes?

retrieval_service/datastore/providers/firestore.py
averikitsch: Ok keeping as strings are fine. I don't think there is any benefit for using a timestamp unless we need easier comparisons.

retrieval_service/datastore/providers/firestore.py
ferdeleong: That is checked inside `delete_indexes` function

retrieval_service/datastore/providers/firestore.py
```diff
@@ -35,47 +38,111 @@ class Config(BaseModel, datastore.AbstractConfig):
...
+    amenities_collection: AsyncQuery
 
     @datastore.classproperty
     def kind(cls):
         return FIRESTORE_IDENTIFIER
 
     def __init__(self, client: AsyncClient):
         self.__client = client
+        self.policies_collection = AsyncQuery(self.__client.collection("policies"))
+        self.amenities_collection = AsyncQuery(self.__client.collection("amenities"))
```
kurtisvg, line 51: these should probably be reduced visibility -- either "__x_collection" or "_x_collection".

retrieval_service/datastore/providers/firestore.py
```diff
@@ -35,47 +38,111 @@ class Config(BaseModel, datastore.AbstractConfig):
 
 class Client(datastore.Client[Config]):
     __client: AsyncClient
+    policies_collection: AsyncQuery
+    amenities_collection: AsyncQuery
```
kurtisvg, line 42: nit: not sure these need to be class properties

retrieval_service/datastore/providers/firestore.py
```diff
@@ -35,47 +38,111 @@ class Config(BaseModel, datastore.AbstractConfig):
...
+            collection_exists = collection_ref.limit(1).stream()
+            if not collection_exists:
+                continue
+
+            docs = collection_ref.stream()
+            async for doc in docs:
+                delete_tasks.append(asyncio.create_task(doc.reference.delete()))
+        await asyncio.gather(*delete_tasks)
+
+    async def delete_indexes(index_list):
```
kurtisvg, line 70: Please add a type:
```suggestion
    async def delete_indexes(index_list):
```
Also nit: consider `indices` as the name instead of including type in variable

retrieval_service/datastore/providers/firestore.py
```diff
@@ -35,47 +38,111 @@ class Config(BaseModel, datastore.AbstractConfig):
...
+                    "composite",
+                    "delete",
+                    index,
+                    "--database=(default)",
+                    "--quiet",  # Added to suppress delete warning
+                ]
+
+                delete_vector_index_process = await asyncio.create_subprocess_exec(
+                    *delete_vector_index,
+                )
```
kurtisvg, line 89: nit: why not the following
```suggestion
                delete_vector_index_process = await asyncio.create_subprocess_exec(
                    "gcloud",
                    "alpha",
                    "firestore",
                    "indexes",
                    "composite",
                    "delete",
                    index,
                    "--database=(default)",
                    "--quiet",  # Added to suppress delete warning
                )
```

retrieval_service/datastore/providers/firestore.py
```diff
@@ -35,47 +38,111 @@ class Config(BaseModel, datastore.AbstractConfig):
...
+        list_vector_index_process = await asyncio.create_subprocess_exec(
+            *list_vector_index,
+            stdout=asyncio.subprocess.PIPE,
+        )
+
+        # Capture output and ignore stderr
+        stdout, __ = await list_vector_index_process.communicate()
+
+        # Decode and format output
+        indexes = stdout.decode().strip().split("\n")
```
kurtisvg, line 132: nit: consider moving this into a function -- it's somewhat complicated and might be better modularize because you can replace with with a better abstract using an API call later

retrieval_service/datastore/providers/firestore.py
```diff
@@ -35,47 +38,111 @@ class Config(BaseModel, datastore.AbstractConfig):
...
+
+        # Check if the indexes already exist; if so, delete indexes
+        # Extract collection and index-id from file path
+        # Assign the index-id to the corresponding collection
+        if indexes != [""]:
+            collections = {"amenities": "", "policies": ""}
+            for line in indexes:
+                collection, index_id = line.split("/")[-3], line.split("/")[-1]
```
kurtisvg, line 140: you should consolidate all of the string parsing in a single place. Doing this will give you more flexibility later to change how you implement this, without this part needing to be rewritten

retrieval_service/requirements.txt
kurtisvg: I still see a git link here?

retrieval_service/datastore/providers/firestore.py
```diff
@@ -326,8 +535,23 @@ async def list_tickets(
...
+        query = collection.find_nearest(
+            vector_field="embedding",
+            query_vector=query_vector,
+            distance_measure=distance_measure,
+            limit=top_k,
+        )
+
+        docs = query.stream()
+        policies = []
+        async for doc in docs:
```
kurtisvg, line 551: nit:
```suggestion
        policies = []
        async for doc in query.stream():
```

retrieval_service/datastore/providers/firestore.py
```diff
@@ -326,8 +535,23 @@ async def list_tickets(
 
     async def policies_search(
         self, query_embedding: list[float], similarity_threshold: float, top_k: int
-    ) -> list[str]:
-        raise NotImplementedError("Semantic search not yet supported in Firestore.")
+    ) -> list[Any]:
+        collection = self.policies_collection
```
kurtisvg, line 539: why the reassignment?

retrieval_service/datastore/providers/firestore.py
```diff
@@ -326,8 +535,23 @@ async def list_tickets(
...
+        query_vector = Vector(query_embedding)
+        distance_measure = DistanceMeasure.DOT_PRODUCT
+        query = collection.find_nearest(
+            vector_field="embedding",
+            query_vector=query_vector,
+            distance_measure=distance_measure,
+            limit=top_k,
+        )
```
kurtisvg, line 547: nit: 
```suggestion
        query = collection.find_nearest(
            vector_field="embedding",
            query_vector=Vector(query_embedding),
            distance_measure=DistanceMeasure.DOT_PRODUCT,
            limit=top_k,
        )
```

kurtisvg: (should apply this in other places too)

retrieval_service/datastore/providers/firestore.py
```diff
@@ -35,47 +38,111 @@ class Config(BaseModel, datastore.AbstractConfig):
...
     def __init__(self, client: AsyncClient):
         self.__client = client
+        self.policies_collection = AsyncQuery(self.__client.collection("policies"))
+        self.amenities_collection = AsyncQuery(self.__client.collection("amenities"))
 
     @classmethod
     async def create(cls, config: Config) -> "Client":
         return cls(AsyncClient(project=config.projectId))
 
+    async def delete_collections(collection_list: list[AsyncCollectionReference]):
```
kurtisvg, line 57: consider reducing visibility 

reducing visibility means fewer things use this, which makes it easier to change later if needed
```suggestion
    async def __delete_collections(collection_list: list[AsyncCollectionReference]):
```

kurtisvg (changes requested):

retrieval_service/datastore/providers/firestore.py
```diff
@@ -42,41 +45,121 @@ def kind(cls):
...
+            "firestore",
+            "indexes",
+            "composite",
+            "list",
+            "--database=(default)",
+            "--format=value(name)",  # prints name field
+        ]
+
+        list_vector_index_process = await asyncio.create_subprocess_exec(
+            *list_vector_index_command,
```
kurtisvg, line 83: nit: consider if it's really useful to put this in it's own array as opposed to just inlining in the command.

retrieval_service/datastore/providers/firestore.py
```diff
@@ -42,41 +45,121 @@ def kind(cls):
...
 
-        # initialize collections
+        # List indexes to check if the collections already exist; if so, delete collections
+        # Assign the index-id to the corresponding collection
+        indexes = await self.__list_vector_index()
+        if indexes != [""]:
+            collections = {"amenities": "", "policies": ""}
+            for line in indexes:
+                collection, index_id = await self.parse_index_info(line)
+                collections[collection] = index_id
```
kurtisvg, line 157: Still some room for improvement here. 

I think you should update `__list_vector_index` to return a dict where `collection_id` -> `index-id`. This means that `__list_vector_index` would call `self.parse_index_info` to parse the line into the right format, include placing the parsed results into a dict. 

The reason for this is that when you are returning a string, you still have additional parsing to do to get something that is useable. By returning the parsed version up front, you return a data type that can be immediately used. It also means that if later on their becomes an API we can call to get indexes, we can just update the 1 place (as opposed to having to update the code to use the API and remove the parse). 

It would simplify this section of code to this (which is considerably more readable):
```
        indexes = await self.__list_vector_index()
        amenities_ref = indexes.get("amenities", "")
        policies_ref = indexes.get("policies", "")
```

you might also want to consider renaming from `__list_vector_index` to `__get_indices` after this change as well.

kurtisvg (approved):

averikitsch (approved):

PR #424, opened 2024-06-25
fix loginUri for llm_demo when original URL has query params

mdunker:
loginUri was adding 'login/google' to the URL, including query params.

This change strips query parameters before concatenating 'login/google'.


mdunker:
Not a conventional commit. Will close and resubmit.

PR #428, opened 2024-06-26
chore: add sql return to retrieval service

Yuan325:
retrieval service to return results and sql query. Sql query will be used to display tracing at the frontend.

This PR only add sql return for postgres and alloydb provider. Other providers will be added in a separate PR.

### DIFF
--- a/retrieval_service/app/routes.py
+++ b/retrieval_service/app/routes.py
@@ -77,7 +77,7 @@ async def get_airport(
             status_code=422,
             detail="Request requires query params: airport id or iata",
         )
-    return {"results": results, "sql": sql, "error": None}
+    return {"results": results, "sql": sql}
 
 
 @routes.get("/airports/search")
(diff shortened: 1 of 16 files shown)


retrieval_service/app/routes.py
```diff
@@ -69,15 +69,15 @@ async def get_airport(
...
     elif iata:
-        results = await ds.get_airport_by_iata(iata)
+        results, sql = await ds.get_airport_by_iata(iata)
     else:
         raise HTTPException(
             status_code=422,
             detail="Request requires query params: airport id or iata",
         )
-    return results
+    return {"results": results, "sql": sql, "error": None}
```
kurtisvg, line 80: Why specify error? It looks like it's always None?

retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -327,36 +327,40 @@ async def export_data(
...
+        self, id: int
+    ) -> tuple[Optional[models.Airport], Optional[str]]:
         async with self.__pool.connect() as conn:
             s = text("""SELECT * FROM airports WHERE id=:id""")
             params = {"id": id}
             result = (await conn.execute(s, params)).mappings().fetchone()
 
         if result is None:
-            return None
+            return None, None
```
kurtisvg, line 339: Why are we returning None? Shouldn't it be the SQL?

kurtisvg (approved):

retrieval_service/app/routes.py
Yuan325: I added this incase we eventually want to display error and improve the frontend error display. I'll remove this for now and add it when we implement that.

PR #429, opened 2024-06-26
fix: use lazy refresh for AlloyDB and Cloud SQL Connector

jackwotherspoon:
The Cloud SQL Python Connector just released a new version `v1.10.0`,
same with AlloyDB Python Connector `v1.2.0` that support setting the
`refresh_strategy` argument of the connector.

Setting the refresh strategy to lazy refresh is recommended for
serverless environments. As the default refresh strategy is to
have background refreshes occur to get the instance metadata and a fresh
certificate to use for the SSL/TLS connection.

However, these background refreshes can be throttled when serverless
environments scale to zero (Cloud Functions, Cloud Run etc.).

This PR will fix the ambiguous errors that are a result of the CPU being
throttled for connectors.

### DIFF
--- a/retrieval_service/datastore/providers/alloydb.py
+++ b/retrieval_service/datastore/providers/alloydb.py
@@ -17,7 +17,7 @@ from datetime import datetime
 from typing import Any, Literal, Optional
 
 import asyncpg
-from google.cloud.alloydb.connector import AsyncConnector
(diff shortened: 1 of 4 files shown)


jackwotherspoon:
@Yuan325 @averikitsch  mind taking a look at this 😄

PR #434, opened 2024-07-08
chore: implement langgraph

Yuan325:
Update implementation to langgraph.

Main changes:
- We create a `UserAgent` for each session in langchain orchestration. For langgraph, we create a single graph and use `config` with `thread_id` to keep track of session. 

This PR does not include the following:
- human-in-the-loop workflow (e.g. book ticket confirmation from user).
- initialize tools with user-specific client. Currently the tools is initialized with a single client (each session/user is supposed to have their own client). Will create a custom ToolNode in another PR so that the Tool is able to get user's `client` object from the state.

### DIFF
--- a/llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
+++ b/llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
@@ -16,22 +16,21 @@ import asyncio
 import os
 import uuid
 from datetime import datetime
-from typing import Any, Dict, List
+from typing import Annotated, Any, Dict, List, Literal, Optional, Sequence, TypedDict
 
 from aiohttp import ClientSession, TCPConnector
 from fastapi import HTTPException
-from langchain.agents import AgentType, initialize_agent
-from langchain.agents.agent import AgentExecutor
 from langchain.globals import set_verbose  # type: ignore
-from langchain.memory import ConversationBufferMemory
-from langchain.prompts.chat import ChatPromptTemplate
-from langchain.tools import StructuredTool
-from langchain_community.chat_message_histories import ChatMessageHistory
 from langchain_core.messages import AIMessage, BaseMessage, HumanMessage
-from langchain_google_vertexai import VertexAI
+from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
+from langchain_core.runnables import RunnableConfig, RunnableLambda
+from langchain_core.tools import StructuredTool
+from langgraph.checkpoint import MemorySaver
+from langgraph.checkpoint.base import empty_checkpoint
 from pytz import timezone
 
 from ..orchestrator import BaseOrchestrator, classproperty
+from .react_graph import create_graph
 from .tools import (
     get_confirmation_needing_tools,
     initialize_tools,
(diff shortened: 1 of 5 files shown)


llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
```diff
@@ -46,131 +49,125 @@
...
+            ):
+                return {
+                    "messages": [
+                        AIMessage(
+                            id=response.id,
+                            content="Sorry, need more steps to process this request.",
+                        )
+                    ]
+                }
+            return {"messages": [response]}
```
kurtisvg, line 100: Why is this function inline? Can we get a comment or explanation as to what it is doing?

llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
```diff
@@ -46,131 +49,125 @@
...
+            """
+            Function to determine which node is called next
+            """
+            messages = state["messages"]
+            last_message = messages[-1]
+            # If the LLM makes a tool call, then we route to the "tools" node
+            if isinstance(last_message, AIMessage) and last_message.tool_calls:
+                return "tools"
+            # Otherwise, we stop (reply to the user)
+            return "end"
```
kurtisvg, line 112: Same with this function -- does it need to be inline?

llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
```diff
@@ -46,131 +49,125 @@
...
+
+        def agent_should_continue(self, state: UserState) -> Literal["tools", "end"]:
+            """
+            Function to determine which node is called next
+            """
+            messages = state["messages"]
+            last_message = messages[-1]
+            # If the LLM makes a tool call, then we route to the "tools" node
+            if isinstance(last_message, AIMessage) and last_message.tool_calls:
+                return "tools"
```
kurtisvg, line 110: Should this be a constant?

llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
```diff
@@ -46,131 +49,125 @@
...
+            # Otherwise, we stop (reply to the user)
+            return "end"
+
+        # define a new graph
+        llm_graph = StateGraph(UserState)
+        llm_graph.add_node("agent", RunnableLambda(call_model))
```
kurtisvg, line 116: should the node strings be constants?

llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
```diff
@@ -46,131 +49,125 @@
...
-                    return {"tool": called_tool.tool, "params": called_tool.tool_input}
-        return None
+        raise NotImplementedError("Irrelevant to LangGraph.")
 
     async def user_session_create(self, session: dict[str, Any]):
         """Create and load an agent executor with tools and LLM."""
-        print("Initializing agent..")
+        client = await self.create_client_session()
+        if self._langgraph_app is None:
+            print("Initializing graph..")
```
kurtisvg, line 141: nit: should really switch to logging at some point. Can file a bug for the backlog though

llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
```diff
@@ -46,131 +49,125 @@
...
 
     def __init__(self):
-        self._user_sessions = {}
+        self._user_sessions = []
 
     @classproperty
     def kind(cls):
         return "langgraph"
 
+    async def create_graph(self, tools, prompt: ChatPromptTemplate, model_name: str):
```
kurtisvg, line 70: should this be an internal function? it's not on the orchestrator interface. Are we calling it publicly?

kurtisvg (changes requested):

llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
Yuan325: This is written similarly to the prebuilt react_agent. Let me know if it's better to move it out (so that this isn't inline).

llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
Yuan325: Updated this. The return will be a string, this string will point towards the next node defined in the conditional edges.

llm_demo/orchestrator/langgraph/langgraph_orchestrator.py
Yuan325: Moved and updated.

kurtisvg (approved):

PR #437, opened 2024-07-16
feat: add ticket validation and insertion to cloudsql postgres

Yuan325:
Also updated alloydb's sqlalchemy syntax for ticket insertion.

### DIFF
--- a/retrieval_service/datastore/providers/cloudsql_postgres.py
+++ b/retrieval_service/datastore/providers/cloudsql_postgres.py
@@ -492,7 +492,29 @@ class Client(datastore.Client[Config]):
         departure_airport: str,
         departure_time: str,
     ) -> Optional[models.Flight]:
-        raise NotImplementedError("Not Implemented")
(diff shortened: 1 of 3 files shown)


retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -506,13 +528,69 @@ async def insert_ticket(
...
+                "user_email": user_email,
+                "airline": airline,
+                "flight_number": flight_number,
+                "departure_airport": departure_airport,
+                "arrival_airport": arrival_airport,
+                "departure_time": departure_time_datetime,
+                "arrival_time": arrival_time_datetime,
+            }
+            result = (await conn.execute(s, params)).mappings()
+            await conn.commit()
```
kurtisvg, line 572: do we need to commit right here?

kurtisvg (approved):
this seems like a feature?

retrieval_service/datastore/providers/cloudsql_postgres.py
Yuan325: Yes, we need commit for inserts.

PR #440, opened 2024-07-18
feat: Implement llm system evaluation

Yuan325:
Implementation of llm system evaluation. 

Usage example: 

If would like to output detailed metric table (row-specific scores), add the following env var `export EXPORT_CSV=True`

Run the following: `python run_evaluation.py`

### DIFF
--- a/llm_demo/evaluation.cloudbuild.yaml
+++ b/llm_demo/evaluation.cloudbuild.yaml
@@ -20,6 +20,15 @@ steps:
   - id: "Run evaluation service"
     name: python:3.11
     dir: llm_demo
+    env: # Set env var expected by tests
+      - "ORCHESTRATION_TYPE=${_ORCHESTRATION_TYPE}"
+      - "RETRIEVAL_EXPERIMENT_NAME=${_RETRIEVAL_EXPERIMENT_NAME}"
+      - "RESPONSE_EXPERIMENT_NAME=${_RESPONSE_EXPERIMENT_NAME}"
     script: |
         #!/usr/bin/env bash
         python run_evaluation.py
+
+substitutions:
+  _ORCHESTRATION_TYPE: "langchain-tools"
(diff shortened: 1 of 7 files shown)


llm_demo/evaluation.cloudbuild.yaml
```diff
@@ -0,0 +1,49 @@
...
+serviceAccount: "projects/$PROJECT_ID/serviceAccounts/evaluation-testing@retrieval-app-testing.iam.gserviceaccount.com" # Necessary for ID token creation
+options:
+  logging: CLOUD_LOGGING_ONLY # Necessary for custom service account
+  dynamic_substitutions: true
+
+substitutions:
+  _ORCHESTRATION_TYPE: "langchain-tools"
+  _RETRIEVAL_EXPERIMENT_NAME: "retrieval-phase-eval-${_PR_NUMBER}"
+  _RESPONSE_EXPERIMENT_NAME:  "response-phase-eval-${_PR_NUMBER}"
+
```
kurtisvg, line 43: Do these still get run on merge to main?

llm_demo/evaluation/eval_golden.py
```diff
@@ -0,0 +1,297 @@
...
+# limitations under the License.
+
+from datetime import datetime, timedelta
+from typing import Any, Dict, List, Optional
+
+from pydantic import BaseModel, Field
+from pytz import timezone
+
+
+class ToolCall(BaseModel):
```
kurtisvg, line 22: nit: some comments on these classes would be nice to have.

What is it representing (or doing)?

kurtisvg (approved):

llm_demo/evaluation.cloudbuild.yaml
Yuan325: Nope, this cloudbuild.yaml only runs on PRs. if we want to run it periodically on main, we can set up a cloud build a replace the name for this.

PR #441, opened 2024-07-18
chore: add cloudbuild and run_evaluation file

Yuan325:
Add run_evaluation.py file and automatic cloudbuild trigger during PR.

Cloudbuild trigger will automatically run the evaluation and the results will be added to GCP under vertexai/experiments.

Detailed metric table (with row specific metrics) could be obtain by running locally with `export EXPORT_CSV=True`.

### DIFF
--- a/llm_demo/evaluation/evaluation.py
+++ b/llm_demo/evaluation/evaluation.py
@@ -39,6 +39,9 @@ async def run_llm_for_eval(
         except Exception as e:
             print(f"error invoking agent: {e}")
         else:
+            eval_data.instruction = (
+                INSTRUCTION + f"\nUser query is '{eval_data.query}'."
+            )
             eval_data.prediction_output = query_response.get("output")
 
             # Retrieve prediction_tool_calls from query response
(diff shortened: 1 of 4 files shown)


llm_demo/evaluation/evaluation.py
```diff
@@ -140,3 +145,23 @@ def evaluate_response_phase(eval_datas: List[EvalData]) -> evaluation_base.EvalR
         experiment=RESPONSE_EXPERIMENT_NAME,
     ).evaluate()
     return eval_result
+
+
+INSTRUCTION = """The Cymbal Air Customer Service Assistant helps customers of Cymbal Air with their travel needs.
```
kurtisvg, line 150: consider how to keep this consistent with the original

kurtisvg (approved):

llm_demo/evaluation/evaluation.py
Yuan325: The only way to extract this is either import the instruction from the lc orchestration file. Else, another way to extract it is using callbacks. Eventually when we use langgraph, this could be extracted via state. :)

PR #446, opened 2024-07-26
chore: add flight booking workflow

Yuan325:
This PR add a new edge and nodes for flight booking workflow. This workflow include requiring human confirmation for flight booking.

Existing:
user ask to book flight -> agent -> insert ticket tool (often times agent might be providing inaccurate information about a flight)

Update:
user ask to book flight -> agent -> validate flight (validate if flight exist and retrieve all the accurate information) -> ask for user confirmation with the flight information -> insert ticket tool

### DIFF
--- a/llm_demo/orchestrator/langgraph/tool_node.py
+++ b/llm_demo/orchestrator/langgraph/tool_node.py
@@ -15,6 +15,7 @@
 import asyncio
 import copy
 import json
+import uuid
 from itertools import repeat
 from typing import Any, Callable, Dict, Optional, Sequence, Union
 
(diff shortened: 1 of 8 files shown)


llm_demo/orchestrator/langgraph/tool_node.py
```diff
@@ -77,8 +78,9 @@ def run_one(call: ToolCall, user_id_token: Optional[str]):
             args = copy.copy(call["args"]) or {}
             args["user_id_token"] = user_id_token
             output = self.tools_by_name[call["name"]].invoke(args, config)
+            tool_call_id = call.get("id") or str(uuid.uuid4())
```
averikitsch, line 81: Why would there not be an id? Can we add a comment on why it's ok to just generate a new UUID?

averikitsch (approved):

llm_demo/orchestrator/langgraph/tool_node.py
Yuan325: Since ToolCall.id is optional, there's a possibility that there's no id. 
ToolMessage.id is required.

PR #448, opened 2024-07-26
chore: update dataset to use dynamic dates

Yuan325:
Update evaluation goldens to use dynamic dates when evaluating query that uses term like "today" or "tomorrow".

### DIFF
--- a/llm_demo/evaluation/eval_golden.py
+++ b/llm_demo/evaluation/eval_golden.py
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from datetime import datetime, timedelta
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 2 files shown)


llm_demo/evaluation/eval_golden.py
```diff
@@ -162,22 +170,22 @@ class EvalData(BaseModel):
     ),
     EvalData(
         category="Insert Ticket",
-        query="I would like to book flight CY 888 departing from SFO on January 1st 2024 at 6am.",
+        query="I would like to book flight CY 922 departing from SFO on 2024-01-01 at 6:38am.",
```
averikitsch, line 173: In the future, it may be good to refactor these so you do not have to manually update the query

averikitsch (approved):

PR #460, opened 2024-08-10
feat: Add category label nodes and BELONGS_TO relationships

ferdeleong:
Add `category` label nodes to the current graph, and connect already existing `amenities` label nodes to them through `BELONGS_TO` relationship. 

Add test cases to:
- Guarantee the successful creation of `category` nodes based on the categories found on the current `amenity_dataset.csv` 
- Guarantee the successful creation of `BELONGS_TO` relationships, one per amenity node based on their category field on the current `amenity_dataset.csv`

### DIFF
--- a/retrieval_service/datastore/providers/neo4j_graph.py
+++ b/retrieval_service/datastore/providers/neo4j_graph.py
@@ -86,6 +86,7 @@ class Client(datastore.Client[Config]):
                 )
 
                 # Create Category node
+                # MERGE prevents duplicate nodes by first checking if they already exist
                 await tx.run(
                     """
                     MERGE (c:Category {name: $category})
@@ -94,6 +95,7 @@ class Client(datastore.Client[Config]):
                 )
 
                 # Create BELONGS_TO relationship
+                # MERGE prevents duplicate relationships by first checking if they already exist
                 await tx.run(
                     """
                     MATCH (a:Amenity {id: $id}), (c:Category {name: $category})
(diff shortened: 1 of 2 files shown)


retrieval_service/datastore/providers/neo4j_graph.py
```diff
@@ -83,6 +85,24 @@ async def create_amenities(tx, amenities):
                     hour=amenity.hour,
                 )
 
+                # Create Category node
+                await tx.run(
+                    """
+                    MERGE (c:Category {name: $category})
```
ferdeleong, line 91: By using merge it ensures if `x category node` already exists, do not create it again.

duwenxin99 (approved):

retrieval_service/datastore/providers/neo4j_graph.py
averikitsch: please add this as a code comment

retrieval_service/datastore/providers/neo4j_graph.py
```diff
@@ -83,6 +85,24 @@ async def create_amenities(tx, amenities):
...
+                    """,
+                    category=amenity.category,
+                )
+
+                # Create BELONGS_TO relationship
+                await tx.run(
+                    """
+                    MATCH (a:Amenity {id: $id}), (c:Category {name: $category})
+                    MERGE (a)-[:BELONGS_TO]->(c)
```
averikitsch, line 100: Should this be MERGE or CREATE?

ferdeleong: Merge is a better option in case csv files have duplicated data.

PR #477, opened 2024-08-23
feat: Add similarity threshold to amenity search

ferdeleong:
Add score to results from vector similarity search and filter by similarity threshold.

Limit vector similarity search results to top k to limit the graph retrieved nodes readability.

Return SIMILAR_TO results in a collection (list) to avoid details from the source node for each retrieved node.

Cleaning csv file from duplicated relationships and add unidentified LLM relationships.


retrieval_service/datastore/providers/neo4j_graph_test.py
```diff
@@ -171,48 +171,58 @@ async def test_total_similar_to_relationships_count(ds: neo4j_graph.Client):
         print(count)
 
     # Bi-directional pairs are counted only once
-    expected_count = 217
+    expected_count = 172
```
ferdeleong, line 174: I polished relationships.csv file, causing the deletion of some irrelevant relationships

retrieval_service/datastore/providers/neo4j_graph.py
```diff
@@ -227,30 +227,34 @@ async def amenities_search(
...
-            # Limit retrieval to 2 target nodes related to the source node, linked by SIMILAR_TO relationship
-            # Lower case relationship due to graph generator output
+            # Retrieve up to 2 source nodes along with all nodes connected by the SIMILAR_TO relationship
+            # Use lowercase relationship names to align with the graph generator's output format
+            # Similarity scores are bounded between 0 and 1, scores closer to 1 indicating higher similarity
             result = await session.run(
                 """
-                CALL db.index.vector.queryNodes('amenity_embedding', $top_k, $query_embedding)
-                YIELD node AS sourceAmenity
+                CALL db.index.vector.queryNodes('amenity_embedding', 2, $query_embedding)
```
averikitsch, line 235: Should we still support top_k instead of hard coding it to 2?

duwenxin99 (approved):

retrieval_service/datastore/providers/neo4j_graph.py
ferdeleong: Any top-k greater than 2 will result in an extensive node list that's unpractical to work with

PR #484, opened 2024-09-10
fix: Reuse connector object across different database connections.

### LINKED ISSUE #416: Inefficient use of connection pools that leverage a `Connector`
The Cloud SQL and AlloyDB datastore providers could improve their use of connection pools, leading to more efficient quota and resource consumption.

https://github.com/GoogleCloudPlatform/genai-databases-retrieval-app/blob/69872805d55eb7332bf533417d44faf2f39b9687/retrieval_service/datastore/providers/cloudsql_postgres.py#L57-L73

The [`creator`](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.creator) argument (same for `async_creator`) for SQLAlchemy is called on each new database connection. This means that initializing a Cloud SQL or AlloyDB `Connector` object within the `get_conn` is creating a new Connector for each connection pool connection and not sharing a single `Connector` across all connections. 

For heavy usage this would lead to Cloud SQL or AlloyDB quotas being exhausted and for applications to begin to error.

A more efficient use would be to create a single `Connector` outside of `get_conn` so that it can be shared across all connections.

anubhav756:
Fixes #416


retrieval_service/datastore/providers/alloydb.py
```diff
@@ -53,19 +54,19 @@ def __init__(self, pool: AsyncEngine):
...
-                conn: asyncpg.Connection = await connector.connect(
-                    # Alloydb instance connection name
-                    f"projects/{config.project}/locations/{config.region}/clusters/{config.cluster}/instances/{config.instance}",
-                    "asyncpg",
-                    user=f"{config.user}",
-                    password=f"{config.password}",
-                    db=f"{config.database}",
-                    ip_type="PUBLIC",
-                )
+            conn: asyncpg.Connection = await cls.__connector.connect(
```
vishwarajanand, line 61: Same here, try moving this block of code to inside the `getconn` method.

`datastore/providers/alloydb.py:61: error: Item "None" of "AsyncConnector | None" has no attribute "connect"  [union-attr]`

```suggestion
            if cls.__connector == None:
                cls.__connector = AsyncConnector(refresh_strategy=RefreshStrategy.LAZY)
            conn: asyncpg.Connection = await cls.__connector.connect(
```

anubhav756: Awesome. Thanks for the suggestion!

retrieval_service/datastore/providers/cloudsql_postgres.py
```diff
@@ -52,20 +53,21 @@ def __init__(self, pool: AsyncEngine):
...
-            ) as connector:
-                conn: asyncpg.Connection = await connector.connect_async(
-                    # Cloud SQL instance connection name
-                    f"{config.project}:{config.region}:{config.instance}",
-                    "asyncpg",
-                    user=f"{config.user}",
-                    password=f"{config.password}",
-                    db=f"{config.database}",
-                )
+            conn: asyncpg.Connection = await cls.__connector.connect_async(
```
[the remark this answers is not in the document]
anubhav756: Thanks for the suggestion!

kurtisvg (approved):

PR #486, opened 2024-09-11
feat: Switch the llm to ChatVertexAI

anubhav756:


llm_demo/orchestrator/langgraph/react_graph.py
```diff
@@ -85,7 +85,7 @@ async def create_graph(
     tool_node = ToolNode(tools)
 
     # model node
-    model = VertexAI(max_output_tokens=512, model_name=model_name, temperature=0.0)
+    model = ChatVertexAI(max_output_tokens=512, model_name=model_name, temperature=0.0)
```
Yuan325, line 88: In order to use tool calling, we will have to add `.bind_tools(tools)`. source: https://python.langchain.com/docs/how_to/tool_calling/

anubhav756: I've added the `.bind_tools(tools)` call for now. We can add other required changes in a future PR.

kurtisvg (approved):

PR #493, opened 2024-09-18
feat: Consolidate postgres providers

anubhav756:
Create a common `PostgresDatastore` class to encapsulate all the DB query logic based on a given connection pool. Reuse the same logic across AlloyDB & CloudSQL Postgres providers for better code maintainability and less redundancy.

### DIFF
--- a/retrieval_service/datastore/providers/alloydb.py
+++ b/retrieval_service/datastore/providers/alloydb.py
@@ -12,20 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import asyncio
-from datetime import datetime
 from typing import Any, Literal, Optional
 
 import asyncpg
 from google.cloud.alloydb.connector import AsyncConnector, RefreshStrategy
 from pgvector.asyncpg import register_vector
 from pydantic import BaseModel
-from sqlalchemy import text
 from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
 
 import models
 
 from .. import datastore
+from .postgres_datastore import PostgresDatastore
 
 ALLOYDB_PG_IDENTIFIER = "alloydb-postgres"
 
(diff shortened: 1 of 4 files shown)


retrieval_service/datastore/providers/alloydb.py
```diff
@@ -532,90 +172,30 @@ async def insert_ticket(
...
-
-        res = [r["content"] for r in results]
-        return res, sql
+        return await self.__pg_ds.policies_search(
+            query_embedding, similarity_threshold, top_k
+        )
 
     async def close(self):
-        await self.__pool.dispose()
+        return await self.__pg_ds.close()
```
vishwarajanand, line 201: Is return needed here?

vishwarajanand (approved):
LGTM, approving to unblock..
I checked the queries moved to `postgres_datastore` seem similar (and do not miss anything obviously) from the queries deleted.

retrieval_service/datastore/providers/alloydb.py
anubhav756: You're right, the potential confusion outweighs the future benefit and consistency with the methods above.  I'll remove the `return` in a separate PR. Thanks!

kurtisvg:
@anubhav756 Let's be careful about enabling automerge when having multiple reviewers

retrieval_service/datastore/providers/cloudsql_postgres_test.py
```diff
@@ -330,7 +330,7 @@ async def test_get_amenity(ds: cloudsql_postgres.Client):
         saturday_end_hour=None,
     )
     assert res == expected
-    assert sql is None
+    assert sql is not None
```
kurtisvg, line 333: This doesn't seem like a particularly useful check -- we should either remove it or assert the SQL is something expected

retrieval_service/datastore/providers/postgres_datastore.py
```diff
@@ -0,0 +1,569 @@
...
+            if not result:
+                raise Exception("Ticket Insertion failure")
+
+    async def list_tickets(
+        self,
+        user_id: str,
+    ) -> tuple[list[Any], Optional[str]]:
+        async with self.__pool.connect() as conn:
+            sql = """
+                SELECT user_name, airline, flight_number, departure_airport, arrival_airport, departure_time, arrival_time FROM tickets
```
kurtisvg, line 534: nit: this is a pretty long line, consider refactoring it onto multiple lines

PR #494, opened 2024-09-18
feat: consolidate Postgres providers

anubhav756:
To align with our existing AlloyDB and CloudSQL providers, we're updating the non-cloud Postgres provider to use SQLAlchemy. This also consolidates AlloyDB and CloudSQL Postgres interactions through the non-Cloud Postgres client by reusing its query handlers.

Also, the `format_sql` helper now accepts `dict` for its `params` argument. This change accommodates the AlloyDB and CloudSQL Postgres providers, and now, post consolidation, non-cloud Postres client as well, which pass `params` as `dict` through `postgres.py`.  The updated `format_sql` helper is used in `postgres.py` query handlers to format SQL queries, populate them with variables, and return them as traces.


vishwarajanand (approved):
LGTM

retrieval_service/datastore/helpers.py
```diff
@@ -12,25 +12,33 @@
...
 
+from typing import Union
+
 import sqlparse
 
 
-def format_sql(sql: str, params):
+def format_sql(sql: str, params: Union[tuple, dict]):
     """
     Format postgres sql to human readable text
```
vishwarajanand, line 22:
```suggestion
    Format Postgres SQL to human-readable text by replacing placeholders.
    Handles tuple-based ($1, $2, ...) and dict-based (:key) formats.
```

vishwarajanand:
I believe this needs to be merged before https://github.com/GoogleCloudPlatform/genai-databases-retrieval-app/pull/493

retrieval_service/datastore/helpers.py
```diff
@@ -12,25 +12,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Union
+
 import sqlparse
 
 
-def format_sql(sql: str, params):
+def format_sql(sql: str, params: Union[tuple, dict]):
```
kurtisvg, line 20: Why can't we just consolidate our use between tuples or dicts?

kurtisvg (changes requested):

retrieval_service/datastore/helpers.py
anubhav756: Thanks for the suggestion. That simplifies the `format_sql`.

kurtisvg: Still looks the same to me?

kurtisvg (changes requested):
PR looks unchanged

retrieval_service/datastore/helpers.py
anubhav756: My bad. I should've clarified that this is already addressed in #495, which consolidates the non-cloud Postgres client. Since that was the only client using `format_sql` with `params` as a `tuple`,  we can now safely remove that functionality. Also agree with your comment to merge these 2 PRs. Thanks!

kurtisvg (approved):

PR #495, opened 2024-09-18
feat: Consolidate non-cloud postgres provider

anubhav756:
To align with our existing AlloyDB and CloudSQL providers, we're updating the non-cloud Postgres provider to use SQLAlchemy. This consolidates all Postgres interactions through the common `postgres_datastore.py` reusing its query handlers.


retrieval_service/datastore/providers/postgres.py
```diff
@@ -497,78 +167,30 @@ async def insert_ticket(
...
+        return await self.__pg_ds.policies_search(
+            query_embedding, similarity_threshold, top_k
         )
 
-        results = [r["content"] for r in results]
-        return results, format_sql(sql, params)
-
     async def close(self):
-        await self.__pool.close()
+        return await self.__pg_ds.close()
```
vishwarajanand, line 196: No need to return

vishwarajanand (approved):
Approving to unblock.. changes LGTM with minor suggestions on variabe names.

retrieval_service/datastore/providers/postgres.py
anubhav756: Thanks for pointing that out.

retrieval_service/datastore/providers/postgres.py
```diff
@@ -39,27 +41,31 @@ class Config(BaseModel, datastore.AbstractConfig):
...
-    __pool: asyncpg.Pool
+    __pg_ds: PostgresDatastore
 
     @datastore.classproperty
     def kind(cls):
         return POSTGRES_IDENTIFIER
 
-    def __init__(self, pool: asyncpg.Pool):
-        self.__pool = pool
+    def __init__(self, pool: AsyncEngine):
```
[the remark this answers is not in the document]
anubhav756: Thanks for the suggestion.

retrieval_service/datastore/providers/alloydb.py
```diff
@@ -47,8 +47,8 @@ class Client(datastore.Client[Config]):
     def kind(cls):
         return ALLOYDB_PG_IDENTIFIER
 
-    def __init__(self, pool: AsyncEngine):
-        self.__pg_ds = PostgresDatastore(pool)
+    def __init__(self, async_engine: AsyncEngine):
```
kurtisvg, line 50: nit: I think "pool" is more understandable (because it's a connection pool) than "async_engine" but maybe that's personal preference

retrieval_service/datastore/providers/postgres.py
```diff
@@ -39,31 +41,35 @@ class Config(BaseModel, datastore.AbstractConfig):
 
 
 class Client(datastore.Client[Config]):
-    __pool: asyncpg.Pool
+    __pg_ds: PostgresDatastore
```
kurtisvg, line 44: nit: you could consider using "postgres.Client" as the "source" class for AlloyDB / Cloud SQL instead of all 3 relying on PostgresDatastore -- this would validate that the source class meets the Client interface

kurtisvg (approved):

retrieval_service/datastore/providers/alloydb.py
anubhav756: Thanks for the feedback! I opted for `async_engine` to emphasize that this class interacts with the database through SQLAlchemy's `AsyncEngine`, which provides a higher level of abstraction, rather than directly with the underlying connection pool.

While "pool" might be more literal, I believe "async_engine" better reflects its functionality.

retrieval_service/datastore/providers/postgres.py
anubhav756: That's a great suggestion! Thanks!

anubhav756: Could you elaborate?

Assuming you meant that changing the parent class from `datastore.Client[Config]` to `postgres.Client` for the `alloydb.Client` and `cloudsql_postgres.Client`, I ran into some interesting type checking issues as follows:

```
datastore/providers/cloudsql_postgres.py:51: error: Argument 1 of "create" is incompatible with supertype "Client"; supertype defines the argument type as "Config"  [override]
datastore/providers/cloudsql_postgres.py:51: note: This violates the Liskov substitution principle
datastore/providers/cloudsql_postgres.py:51: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
datastore/providers/alloydb.py:51: error: Argument 1 of "create" is incompatible with supertype "Client"; supertype defines the argument type as "Config"  [override]
datastore/providers/alloydb.py:51: note: This violates the Liskov substitution principle
datastore/providers/alloydb.py:51: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

retrieval_service/datastore/helpers.py
```diff
@@ -12,25 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Union
+
 import sqlparse
 
 
-def format_sql(sql: str, params):
+def format_sql(sql: str, params: dict):
```
kurtisvg, line 20: I see you made the changes here -- do you want to just merge this into a single PR?

retrieval_service/datastore/providers/postgres.py
kurtisvg: Right now you are doing this:

```mermaid
flowchart LR
    A[AlloyDBClient] --composed by--> D[PostgresDatastore]
    C[CloudSQLPGClient] --composed by--> D
    P[PostgresClient] --composed by--> D
```

I'm just suggesting you get rid of one wrapper class so it looks like this:
```mermaid
flowchart LR
    A[AlloyDBClient] --composed by--> P[PostgresClient]
    C[CloudSQLPGClient] --composed by--> P
```

Which you would implement with something like:
```python

    def __init__(self, pgclient: PostgresClient): 
        self.__pgclient = pgclient

    @classmethod
    async def create(cls, config: Config) -> "Client":
         # connection pool gobble gook 
        return cls(PostgresClient(async_engine))
```

retrieval_service/datastore/helpers.py
anubhav756: Thanks for the suggestion. Merging this PR into #494.

PR #508, opened 2024-11-15
fix: Add test cases to improve coverage for postgres and over more tools.

anubhav756:


retrieval_service/datastore/providers/postgres_test.py
```diff
@@ -604,3 +604,26 @@ async def test_policies_search(
...
+async def test_validate_ticket_valid(ds: postgres.Client):
+    flight, sql = await ds.validate_ticket(
+        airline="UA",
+        flight_number="1158",
+        departure_airport="SFO",
+        departure_time="2024-01-01 05:57:00",
+    )
+    assert flight is not None
+    assert sql is not None
+    assert flight.id == 1
```
Yuan325, line 618: can we mock the whole flight (similar to the other test cases)?

retrieval_service/datastore/providers/postgres_test.py
```diff
@@ -604,3 +604,26 @@ async def test_policies_search(
...
+
+
+async def test_validate_ticket_invalid(ds: postgres.Client):
```
Yuan325, line 621: We can put both valid and invalid under the same function (similar to the other test cases to keep it consistent) while using `@pytest.mark.parametrize` annotation.

Yuan325, line 621: We can put both valid and invalid under the same function (similar to the other test cases to keep it consistent) while using `@pytest.mark.parametrize` annotation.

Yuan325 (changes requested):

Yuan325:
While you're on this, can you also add the test_validate_ticket test case to other datastore as well? 

And if the test coverage increase, can you increase the limit under retrieval_app/coverage (so that it will throw an error if more code are added in the future without providing test cases).

retrieval_service/app/app_test.py
```diff
@@ -783,3 +808,80 @@ def test_policies_search(m_datastore, app, method_name, params, mock_return, exp
...
+            json={
+                "airline": "CY",
+                "flight_number": "888",
+                "departure_airport": "LAX",
+                "arrival_airport": "JFK",
+                "departure_time": "2024-01-01 08:08:08",
+                "arrival_time": "2024-01-01 08:08:08",
+            },
+        )
+        assert response.status_code == 422
```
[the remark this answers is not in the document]
Yuan325: And is it possible to mock a success insert_ticket operation?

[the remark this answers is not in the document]
anubhav756: > For this, it should be throwing a different error (similar to the list flight below):
> 
> ```
> raise HTTPException(
>             status_code=401,
>             detail="User login required for data insertion",
>         )
> ```
> 
> Will this work if we don't mock it at all? Since the header won't have the User-Id-Token field, so the retrieval of id_token will fail.

Thanks for pointing that out. Fixed!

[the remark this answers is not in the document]
anubhav756: > And is it possible to mock a success insert_ticket operation?

That's a good idea. Added :)

retrieval_service/app/app_test.py
```diff
@@ -783,3 +808,80 @@ def test_policies_search(m_datastore, app, method_name, params, mock_return, exp
...
+                "url": "https://errors.pydantic.dev/2.6/v/missing",
+            },
+        ]
+
+
+@patch.object(id_token, "verify_oauth2_token")
+@patch.object(datastore, "create")
+def test_list_tickets_missing_user_info(m_datastore, m_verify_oauth2_token, app):
+    m_datastore = AsyncMock()
+    m_verify_oauth2_token.side_effect = ValueError("invalid token")
```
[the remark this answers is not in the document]
anubhav756: In this particular case (ie. failing case), no, since it would anyway fail with the dummy value. However, since I parameterised these tests, and since they now also include the success case, we do need to mock this now with the parameter provided. Note that still for the failure case, I just return `None`, and I still don't mock the side effect, since that would fail as intended. So thanks for the suggestion! That definitely simplifies 🙂

retrieval_service/datastore/providers/postgres_test.py
anubhav756: Good idea. Done.

retrieval_service/datastore/providers/postgres_test.py
anubhav756: Thanks for the great suggestion! Done.

anubhav756:
> While you're on this, can you also add the test_validate_ticket test case to other datastore as well?
> 
> And if the test coverage increase, can you increase the limit under retrieval_app/coverage (so that it will throw an error if more code are added in the future without providing test cases).
> 
> Thank you! This is great~

Done! Thanks for all the suggestions!

Yuan325 (approved):
Thank you for improving our tests!

PR #511, opened 2024-11-18
fix: Fix spanner providers issue with timezones while validating ticket.

anubhav756:


anubhav756:
This change is included in #508 since both the PRs are required to keep the presubmits green.

PR #535, opened 2025-05-21
chore: update model

vishwarajanand:
Updates model name since `gemini-pro` seems deprecated now.


anubhav756 (approved):
Thanks for the update!

PR #540, opened 2025-06-20
feat: Consolidate on a single LangGraph orchestration

anubhav756:
This PR removes LangChain tools and Vertex AI Function Calling orchestration and consolidates on LangGraph.

It also flattens the directory structure and refactors to simplify some parts of code for easier understanding.


.github/workflows/lint.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        dir: [retrieval_service, llm_demo]
```
Yuan325, line 37: might want to check if we are running lint on root dir. (for both this and the fallback lint).

Also wanna make sure that we are running it in `/evaluation` and `/orchestrator` as well.

app.py
```diff
@@ -27,7 +27,7 @@
 from markdown import markdown
 from starlette.middleware.sessions import SessionMiddleware
 
-from orchestrator import createOrchestrator
+from orchestrator import Orchestrator
```
Yuan325, line 30: Feels like the name `orchestrator` is very broad and unclear. Perhaps something like `agent` might be better? WDYT? and @kurtisvg any preference? 

same for the `orchestrator` folder name.

orchestrator/orchestrator.py
```diff
@@ -42,6 +42,8 @@
 
 
 class LangGraphOrchestrator(BaseOrchestrator):
```
Yuan325, line 44: This might throw an error? Seems like the `BaseOrchestrator` is missing.

But regardless I think we can just remove the `BaseOrchestrator` overall and just implement a `class LangGraphAgent` or `LangGraphGraph` directly without using the interface.

run_evaluation.py
```diff
@@ -26,7 +26,7 @@
     goldens,
     run_llm_for_eval,
 )
-from orchestrator import createOrchestrator
+from orchestrator import Orchestrator
```
Yuan325, line 29: Where is the `Orchestrator()` defined? I don't see that in `orchestrator/orchestrator.py`

evaluation.cloudbuild.yaml
```diff
@@ -14,14 +14,11 @@
 steps:
```
Yuan325, line 14: Might want to stop running the evaluation at cloud build. This will probably fail it since the evaluation is built on top of the langchain orchestration.

Yuan325 (changes requested):
Do we want to just make this whole thing cleaner by also removing the `/llm_demo` and `/retrieval_app`? It's hard to track what files will be left after the change since it still includes all other stuffs.

Merging all the cymbal_air_updates to a separate branch first might be a good idea. That way we can merge all changes at once to `main` when everything is done.

.github/workflows/lint.yml
anubhav756: Yes, in fact the entire `strategy` section is removed in [#543](https://github.com/GoogleCloudPlatform/genai-databases-retrieval-app/pull/543):

I assume that since we would remove the `strategy` section, we would also cover all the files in root directory (including evaluation and orchestrator as well). WDYT? 🙂 

image

orchestrator/orchestrator.py
anubhav756: Yes, clearing those parts is covered in #542.

app.py
anubhav756: Added. Thanks for the suggestion!

run_evaluation.py
anubhav756: #542 adds those changes. Maybe I should include this change in #542 itself. 🤔

PR #546, opened 2025-06-23
feat: Update text embedding model to gemini-embedding-001

anubhav756:


Yuan325 (commented):
This PR might need to be rebased. And i thought we had remove the `llm_demo` directory?

anubhav756:
> This PR might need to be rebased. And i thought we had remove the `llm_demo` directory?

Rebased. Yes, the `llm_demo` is removed in the other streamlining PRs. Would you suggest placing this PR after other streamlining PRs?

llm_demo/app.py
```diff
@@ -32,6 +32,8 @@
 routes = APIRouter()
 templates = Jinja2Templates(directory="templates")
 
+EMBEDDING_MODEL_NAME = "gemini-embedding-001"
```
Yuan325, line 35: not sure if we want to add it here. I don't think the `llm_demo` uses the embedding model at all. And iirc, we are removing this folder right?

run_generate_embeddings.py
```diff
@@ -0,0 +1,91 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+import csv
+from langchain_google_vertexai import VertexAIEmbeddings
+import models
+from app import EMBEDDING_MODEL_NAME
```
Yuan325, line 19: I don't think this will be grabbing from app anymore once we move it out?

Yuan325 (commented):
Is it worth moving all the `run_generate_embeddings.py` and `run_generate_policy_dataset.py` etc. into root dir? Feels like it will get cluttered really quickly. iirc we will also have a `run_database_init.py` and `run_database_export.py` eventually. 

Might be worth to move these into a folder (or even the `/data` dir?). With that, we can have a separate file `embedding_model.py` with a single line for the embedding model. WDYT?

anubhav756:
> Is it worth moving all the `run_generate_embeddings.py` and `run_generate_policy_dataset.py` etc. into root dir? Feels like it will get cluttered really quickly. iirc we will also have a `run_database_init.py` and `run_database_export.py` eventually.
> 
> Might be worth to move these into a folder (or even the `/data` dir?). With that, we can have a separate file `embedding_model.py` with a single line for the embedding model. WDYT?

That's a great idea now that we have a more flat directory structure, vs the separate `llm_demo/` and `retrieval_service` that provided some segregation. I've moved these to the `data/` directory. Thanks for pointing that out!

llm_demo/app.py
anubhav756: Thanks for the suggestion. I originally kept it here as I was thinking about generating user query embeddings (for comaparison against the vector embeddings of datasource) within the app itself, which would have required the embedding model. However, I'm now planning to do this within the database itself. Going forward, we'll migrate to the Semantic Search Tool when it's available. I've moved it to `data/` directory.

models/models.py
```diff
@@ -0,0 +1,131 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import ast
+import datetime
+from typing import Optional
+
+from pydantic import BaseModel, ConfigDict, field_validator
+
+EMBEDDING_MODEL_NAME = "gemini-embedding-001"
```
Yuan325, line 21: Probably don't want to add the `embedding model name` in this file. iiuc, the `model` in this file meant pydantic model, it's different from the model we meant when we say embedding model.

Yuan325 (changes requested):
Final nit, ideally we don't want to put the embedding model name in the models.py? I think they are referring to a different "model".

anubhav756:
> Final nit, ideally we don't want to put the embedding model name in the models.py? I think they are referring to a different "model".

I think I forgot to move this one while removing the eval module. Thanks for pointing that out!

data/run_generate_policy_dataset.py
```diff
@@ -0,0 +1,111 @@
...
+import time
+
+import pandas as pd
+from langchain_google_vertexai import VertexAIEmbeddings
+from langchain_text_splitters import (
+    MarkdownHeaderTextSplitter,
+    RecursiveCharacterTextSplitter,
+)
+
+from data.run_generate_embeddings import EMBEDDING_MODEL_NAME
```
Yuan325, line 24:
```suggestion
from .run_generate_embeddings import EMBEDDING_MODEL_NAME
```
last nit: Will this work?

Yuan325 (approved):

PR #547, opened 2025-06-26
feat: Deploy and connect Toolbox service in Cloud Run

anubhav756:
## Summary
This PR updates the Cloud Build configuration to automatically deploy a Toolbox service on Cloud Run and securely connect it to the main application.

## Changes

* The `integration.cloudbuild.yaml` for our integration environment now includes steps to deploy the Toolbox service.
* The deployed Toolbox service's URL is passed as an environment variable to the main application.
* Communication between the main app and the Toolbox service is secured using service account-based authentication.
  * The `ToolboxClient` now fetches the current service account's ID token and includes it in the `client_headers` for all outgoing requests to the Toolbox service.


agent/tools.py
```diff
@@ -13,17 +13,34 @@
...
+            "Ensure that the GOOGLE_APPLICATION_CREDENTIALS environment variable is set correctly.",
+            UserWarning,
+        )
+        return None
+
+
 # Tools for agent
 async def initialize_tools():
-    client = ToolboxClient(TOOLBOX_URL)
+    client = ToolboxClient(TOOLBOX_URL, client_headers=__get_client_headers())
```
Yuan325, line 43: I thought we are using langgraph configuration for this?

integration.cloudbuild.yaml
```diff
@@ -12,24 +12,66 @@
 # limitations under the License.
 
 steps:
-  - id: "Deploy to Cloud Run"
+  # Step 1: Deploy the Toolbox service from a public image and capture its URL
```
Yuan325, line 15: Might take a while to deploy cloud run every time we run the tests? 

I'm wondering if it's better to pre-deploy Toolbox cloud run? And add a trigger to re-run the cloud run if our `tools.yml` is updated. similar to how retrieval_service is pre-deployed in cloud run previously.

integration.cloudbuild.yaml
```diff
@@ -12,24 +12,66 @@
...
+        set -e
+
+        # Check if the URL file from the previous step exists and is not empty
+        if [ ! -s /workspace/toolbox_url.txt ]; then
+          echo "ERROR: /workspace/toolbox_url.txt is missing or empty."
+          exit 1
+        fi
+
+        # For debugging, print the URL we are about to use
+        echo "Deploying Frontend with TOOLBOX_URL=$$(cat /workspace/toolbox_url.txt)"
```
Yuan325, line 54: we can also pre-set the `TOOLBOX_URL` link as an env var?

Yuan325 (changes requested):

integration.cloudbuild.yaml
anubhav756: Yea, previously I had set it up that way itself. However, I thought of deploying dynamically as it would isolate the tests well. However, on a second thought, it might not be necessary, given our tools file for integration tests are fairly constant, and the amount of time it saves. I've reverted to the pre-deployed service approach for toolbox and injected its URL to the app. Thanks for the suggestion!

agent/tools.py
```diff
@@ -13,17 +13,45 @@
...
+
+    This function uses the application's default credentials to generate an
+    identity token.
+
+    Returns:
+        Optional[dict[str, str]]: A dictionary containing the Authorization
+        header if authentication is successful, otherwise None.
+    """
+    try:
+        id_token = fetch_id_token(Request(), TOOLBOX_URL)
```
Yuan325, line 41: Is this specific towards Cloud Run? What if user runs this locally?

Also, curious why are we grabbing Toolbox's URL? afaik, we don't have login for toolbox...? I thought we want to grab client from the app?

integration.cloudbuild.yaml
```diff
@@ -74,3 +82,4 @@ substitutions:
   _GCR_HOSTNAME: ${_REGION}-docker.pkg.dev
   _SERVICE: demo-service-${BUILD_ID}
   _REGION: us-central1
+  _TOOLBOX_URL: https://toolbox-service-706535509072.us-central1.run.app
```
Yuan325, line 85:
```suggestion
  _TOOLBOX_URL: https://toolbox-service-${PROJECT_NUMBER}.us-central1.run.app
```

Or we can do `_TOOLBOX_URL: http://127.0.0.1:5000`, and then add a env var substitution within the cloud build trigger.

Yuan325 (changes requested):

agent/tools.py
anubhav756: > Is this specific towards Cloud Run?

Absolutely, this is only specific towards Cloud Run.

> What if user runs this locally?

Then the `except` case logs a warning, and the user's ADC would be used from the env. I can remove the warning if that makes sense, but I kept this around assuming it would still be beneficial to inform them about the `GOOGLE_APPLICATION_CREDENTIALS` env var that's hinted in the warning message, and they get some more hints in the logs while debugging any potential issues in case their ADC is not setup correctly.

> Also, curious why are we grabbing Toolbox's URL?

I think I had added the wrong audience. Fixed it by adding the client ID. Thanks for pointing that out!

PR #550, opened 2025-06-29
fix: Make human-in-the-loop more consistent between graph state and chat history

anubhav756:
## Summary
This PR addresses inconsistencies between the UI display and the agent's internal LangGraph state during the ticket booking process. Previously, hardcoded messages spread across the UI, the backend request handlers, and agent nodes, led to a disjointed UX where chat history, user prompts, and agent responses could become out of sync, particularly after a page refresh.

The core of this change is the introduction of a single, internal helper function (__booking_handler) that centralizes the logic for both confirming and declining a ticket booking. This refactoring ensures that all UI components and chat history entries are a direct reflection of the agent's state.

## Changes
* Created a common `__booking_handler` to manage both booking "accept" and "decline" actions, eliminating code duplication and streamlining maintenance.

* Replaced hardcoded UI text for confirmations and responses with messages sourced directly from the agent's graph state. This guarantees consistency between what the user sees and the agent's conversational history.

* The user's choice (e.g., `"Looks good to me. Book it!"`) and the agent's final response are now reliably added to the user session's chat history for both success and decline scenarios.

* Added a crucial comment explaining the rationale for injecting the "decline" message to the langgraph state *before* invoking the agent.

* Removed unnecessary `params` from the request handlers for better code hygiene.


static/index.js
```diff
@@ -169,7 +170,7 @@ function buildConfirmation(confirmation, messageId) {
         const message = `<div class="chat-bubble ai" id="${message_id}">
         <div class="sender-icon"><img src="static/logo.png"></div>
         <div class="ticket-confirmation">
-            Please confirm the details below to complete your booking
+            ${output}
```
Yuan325, line 173: Will the `${output}` include instructions? Where do we set it/what's the expected output here?

app.py
```diff
@@ -155,56 +155,71 @@ async def chat_handler(request: Request, prompt: str = Body(embed=True)):
...
     confirmation = response.get("confirmation")
     trace = response.get("trace")
+    request.session["history"].append({"type": "ai", "data": {"content": output}})
     # Return assistant response
     if confirmation:
         return json.dumps(
-            {"type": "confirmation", "content": confirmation, "trace": trace}
+            {
+                "type": "confirmation",
+                "content": {"output": output, **confirmation},
```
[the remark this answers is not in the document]
anubhav756: This is needed for building the confirmation dialog in the UI. Specifically, this adds the `tool` and `params` to the response, which is used in the UI to satisfy the `if` condition, start building the confirmation window, and get params to populate in it.
image

static/index.js
anubhav756: > Will the `${output}` include instructions?
No. Just the hardcoded confirmation message.

> Where do we set it
We set it as the graph's response in `user_session_invoke` here:
image

> what's the expected output here?
This is the hardcoded message asking whether the user wants to continue with the booking or not. Previously, the booking confirmation message was hardcoded in multiple locations, creating inconsistencies. This change centralizes the message within its corresponding graph node. The message is now propagated to the UI from this single source, ensuring consistency and maintainability:
image

Yuan325 (approved):

PR #553, opened 2025-07-04
feat: Write scripts that initialize/export database through Toolbox prebuilt tools

anubhav756:


Yuan325 (approved):
Just to confirm, we are only going to support for postgres db right now?

anubhav756:
> Just to confirm, we are only going to support for postgres db right now?

I am starting to test with AlloyDB and CloudSQL PG for now, but will test with other datasources going forward and accommodate necessary changes as required. Thanks for brining this up!

PR #560, opened 2025-08-07
feat: Implement proactive authentication flow

anubhav756:
# Overview
This PR introduces a multi-step validation flow into the agent's graph to handle tools that require authentication. The primary goal is to prevent the agent from attempting to use authenticated tools without proper authentication.

# New Flow Diagram
Untitled diagram _ Mermaid Chart-2025-08-08-210230

# Problem
Previously, the agent would attempt to execute any tool the model chose, without proactively checking for prerequisites. This led to tools requiring a logged-in user fail late in the process, resulting in a poor UX and wasted processing.

# Solution
This PR implements a new conditional logic flow within the graph to address this issue.

## Proactive Auth Check
A new conditional edge has been added to check if a user is logged in before any tool requiring authentication is called.
* An `__is_logged_in` helper function checks the `RunnableConfig` for valid credentials.
* A `request_login_node` was added to inform the user when they need to sign in.
* The graph's conditional edges from the `AGENT_NODE` now route through this authentication check first.

## Prioritized and Sequential Routing Logic
The updated routing logic in `agent_should_continue` ensures a strict order of operations for tools requiring both authentication and confirmation, like `insert_ticket`.
* If no tool is called, the agent responds and ends.
* The graph always checks if any selected tool requires authentication. If the user is not logged in, the flow is immediately directed to `request_login_node` and stops.
* Only after authentication is confirmed does the graph check if the tool requires user confirmation.
* If none of the above conditions is met, the graph proceeds to the standard `tool_node`.

### DIFF
--- a/agent/tools.py
+++ b/agent/tools.py
@@ -39,6 +39,7 @@ async def initialize_tools():
 def get_confirmation_needing_tools():
     return ["insert_ticket"]
 
+
 def get_auth_tools():
     return [
         "insert_ticket",


agent/react_graph.py
```diff
@@ -95,12 +112,15 @@ async def create_graph(
...
-        End --> [*]
+        [*]
+         └──> Agent
+                 ├──(No tool calls)──> [*]
+                 └──(Tool calls)──> Request Login?
+                                                ├──(No)──> Tools ──> Agent
+                                                ├──(Yes, but not logged in)──> [*]
+                                                └──(Yes, and logged in)──> Needs Confirmation?
+                                                                                            ├──(No)──> Tools ──> Agent
+                                                                                            └──(Yes)──> Booking Validation ──> Insert Ticket ──> [*]
```
Yuan325, line 123: Is it possible to reduce the intendations here? might make it tough to read on screen when the line is too long (overflow to the next row).

agent/react_graph.py
```diff
@@ -169,25 +189,45 @@ async def acall_model(state: UserState, config: RunnableConfig):
...
+            return "end"
+
+        # Next, check if any tool requires authentication.
+        for tool_call in last_message.tool_calls:
+            if tool_call["name"] in get_auth_tools():
+                if not __is_logged_in(config):
+                    return "request_login"
+
+        # If authentication passes, then check if any tool needs user confirmation.
+        for tool_call in last_message.tool_calls:
```
Yuan325, line 224:
```suggestion
                    return "request_login"
```

Might be able to check both in the same for loop?

agent/react_graph.py
```diff
@@ -283,6 +327,7 @@ async def insert_ticket_node(state: UserState, config: RunnableConfig):
         {"continue": INSERT_TICKET_NODE, "agent": AGENT_NODE},
     )
     llm_graph.add_edge(INSERT_TICKET_NODE, END)
+    llm_graph.add_edge(REQUEST_LOGIN_NODE, END)
```
Yuan325, line 330: Do we want to end the graph after request login? Or should we continue with the tool call process after request login?

Yuan325 (changes requested):

agent/react_graph.py
anubhav756: Thanks for the suggestion. Made the graph much more compact to make it more vertical than horizontal.

agent/react_graph.py
Yuan325: ahh make sense, yeah i think we start a new thread when user log in.

Yuan325 (approved):
yay!

PR #597, opened 2025-12-15
feat(eval): add LangGraph-focused evaluation

SaxenaAnushka102:
## Description

This commit refactors the evaluation module to primarily support LangGraph-based agents.

Key changes include:
- Adapting `evaluation.py` and `run_evaluation.py` for a LangGraph orchestration type.
- Introducing `evaluate_retrieval_trajectory_phase` function stub in `evaluation.py` for future multi-step tool call evaluation.
- Updating tool names and arguments in the `eval_golden.py` dataset.


evaluation/__init__.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2024 Google LLC
```
anubhav756: Since we're creating this file in 2025, we want to change the license to match that (here and all other files as well).

PS: Doesn't apply if we're updating a file though.

SaxenaAnushka102: Oh okay, got you point! I'll update the license

SaxenaAnushka102:
Perhaps we can only trigger the full evals in presubmit if files within the agent/ or evaluation/ directories are changed. Would that be a good approach?

anubhav756:
gcbrun

PR #657, opened 2026-07-30
fix: add error messages to chat history and graph state

### LINKED ISSUE #656: Add error messages to chat history and graph state for consistency
Currently in case an error occurs in Cymbal Air, the UI shows a "Sorry..." message, but when the user refreshes the page, it goes away. It is also not added to the graph's state.

This could confuse the LLM on further chat invocations.

To fix this we could add the error messages consistently to both the graph's state and chat history for consistency.

emagda97:
Fixes #656

### DIFF
--- a/agent/agent.py
+++ b/agent/agent.py
@@ -52,7 +52,7 @@ class Agent:
         return uuid in self._user_sessions
 
     async def user_session_insert_ticket(self, uuid: str) -> Any:
-        return await self.user_session_invoke(uuid, None)
+        return await self.user_session_invoke(uuid, None, booking_type="confirm")
 
     async def user_session_decline_ticket(self, uuid: str) -> dict[str, Any]:
         config = self.get_config(uuid)
@@ -67,7 +67,7 @@ class Agent:
             content="I changed my mind. Decline ticket booking."
         )
         self._langgraph_app.update_state(config, {"messages": [human_message]})
-        return await self.user_session_invoke(uuid, None)
+        return await self.user_session_invoke(uuid, None, booking_type="decline")
 
     async def user_session_create(self, session: dict[str, Any]):
         """Create and load an agent executor with tools and LLM."""
@@ -101,7 +101,7 @@ class Agent:
         self._user_sessions[session_id] = ""
 
     async def user_session_invoke(
-        self, uuid: str, user_prompt: Optional[str]
+        self, uuid: str, user_prompt: Optional[str], booking_type: Optional[str] = None
     ) -> dict[str, Any]:
         config = self.get_config(uuid)
         cur_message_index = (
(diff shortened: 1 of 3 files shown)


agent/agent.py
```diff
@@ -112,10 +112,41 @@ async def user_session_invoke(
...
+            )
+        except Exception as e:
+            print(f"Error invoking LangGraph: {e}")
+
+            if user_prompt:
+                error_message = "Sorry, we couldn't answer your question 😢"
+                messages_to_add = [
+                    HumanMessage(content=user_prompt),
+                    AIMessage(content=error_message),
+                ]
```
anubhav756, line 129: If an exception occurs during multi-step graph execution (for example, the model successfully decides to call a tool, the tool executes, and then the follow-up model call fails due to an API issue or rate limit), `HumanMessage(content=user_prompt)` and intermediate `ToolMessage`s have already been committed to the checkpoint in earlier steps.

Unconditionally prepending `HumanMessage(content=user_prompt)` here will duplicate the user's message in the graph's history. Furthermore, if an error happens when tool calls are pending, inserting a new `HumanMessage` without corresponding `ToolMessage`s can trigger a 400 error from Gemini/Vertex AI on the next turn.

Consider checking whether `HumanMessage` is already recorded in the current state before adding it?

agent/agent.py
```diff
@@ -112,10 +112,41 @@ async def user_session_invoke(
...
+            elif booking_type == "confirm":
+                error_message = "Sorry, flight booking failed. 😢"
+                messages_to_add = [
+                    HumanMessage(content="Looks good to me. Book it!"),
+                    AIMessage(content=error_message),
+                ]
```
anubhav756, line 135: The graph compiles with `interrupt_after=["booking_validation"]`. When `user_session_insert_ticket` is called, the graph resumes to run `insert_ticket_node`. 

If `__insert_ticket.ainvoke()` throws an exception and we catch it here in `user_session_invoke`, calling `self._langgraph_app.update_state` adds the messages, but it leaves the graph interrupted midway rather than allowing it to finish cleanly at `END`. IIUC, on the user's next turn, LangGraph will attempt to resume the interrupted thread from `booking_validation` rather than starting fresh from `agent`?

Should we catch tool exceptions inside [`insert_ticket_node`](agent/react_graph.py#L260) (just like `tool_node` in `react_graph.py` does?

agent/agent.py
```diff
@@ -101,7 +101,7 @@ async def user_session_create(self, session: dict[str, Any]):
         self._user_sessions[session_id] = ""
 
     async def user_session_invoke(
-        self, uuid: str, user_prompt: Optional[str]
+        self, uuid: str, user_prompt: Optional[str], booking_type: Optional[str] = None
```
anubhav756, line 104: `user_session_invoke` is the generic entry point for interacting with the LangGraph runner. Adding `booking_type: Optional[str] = None` and branching on `"confirm"` and `"decline"` seems to tightly couple general graph execution to ticket booking logic. Should we consider passing an optional `error_fallback_message: Optional[str] = None` or handling fallbacks directly in `user_session_insert_ticket` and `user_session_decline_ticket`?

agent/agent.py
```diff
@@ -112,10 +112,41 @@ async def user_session_invoke(
...
-            config=config,
-        )
+        try:
+
+            final_state = await self._langgraph_app.ainvoke(
+                app_input,
+                config=config,
+            )
+        except Exception as e:
+            print(f"Error invoking LangGraph: {e}")
```
anubhav756, line 122: Using `print()` could lose the exception traceback. Let's replace this with `logging.exception("Error invoking LangGraph")` (or `logging.error(..., exc_info=True)`) to make debugging production issues easier?

app_test.py
```diff
@@ -12,6 +12,198 @@
...
+        response = self.client.get("/")
+        self.assertEqual(response.status_code, 200)
+
+        chat_response = self.client.post("/chat", json={"prompt": "test question"})
+        self.assertEqual(chat_response.status_code, 200)
+
+        data = chat_response.json()
+        self.assertEqual(data["type"], "message")
+        self.assertIn("Sorry, we couldn't answer your question", data["content"])
+
```
anubhav756, line 156: optional: #656 says that when an error occurs, the message displayed in the UI disappears if the user refreshes the page.

Should we add a reload check to assert that reloading the page preserves the history?

anubhav756 (changes requested):

PR #664, opened 2026-08-14
feat: use semantic search from mcp-toolbox

### LINKED ISSUE #660: Implement semantic search
This task is about implementing the Semantic Search tool ([blog](https://medium.com/google-cloud/building-semantic-search-into-your-ai-agents-d72349496340)) in Cymbal air.

For context, currently there are 2 specific tools that use vector search ([`search_amenities`](https://github.com/GoogleCloudPlatform/cymbal-air-toolbox-demo/blob/34913fb55c40f270abfbbc45ade5182f9db99f8c/tools.yaml#L94) and [`search_policies`](https://github.com/GoogleCloudPlatform/cymbal-air-toolbox-demo/blob/34913fb55c40f270abfbbc45ade5182f9db99f8c/tools.yaml#L115)).

At the time of its [implementation](https://github.com/GoogleCloudPlatform/cymbal-air-toolbox-demo/pull/558) we didn't have semantic search implemented within Toolbox, so we had to rely on the Databases for this. Moreover, at that time only AlloyDB supported Vector Searching in production, so we had added only AlloyDB as a datasource for the app.

Now, although, there might be other databases that must be supporting Vector search in production as well, but we have this feature available from within Toolbox, so it makes sense make Cymbal Air to use Toolbox's feature itself. Moreover, it would be better since we could potentially be supporting more databases for vector search rather than relying on them for the feature.

emagda97:


data/run_database_init.py
```diff
@@ -100,9 +100,6 @@ async def initialize_data(
         await execute_sql(f"""INSERT INTO airports VALUES {", ".join(values)}""")
         print("Airports table initialized")
 
-        await execute_sql(
-            "CREATE EXTENSION IF NOT EXISTS google_ml_integration CASCADE"
-        )
```
anubhav756, line 105: Are we removing this because it's only required once, and doesn't need to be run everytime? If so, we might keep it just in case, and also for guiding someone setting up in their own database?

tools.yaml
```diff
@@ -12,6 +12,12 @@ authServices:
   my_google_service:
     kind: google
     clientId: 706535509072-qa5v22ur8ik8o513b0538ufo0ne9jfn5.apps.googleusercontent.com
+embeddingModels:
+  gemini_embedding:
+    kind: gemini
+    model: gemini-embedding-001
```
anubhav756: Don't we have `gemini-embedding-2` now? https://ai.google.dev/gemini-api/docs/embeddings

anubhav756 (changes requested):
nit: In case you agree to change the embedding model, then we might also want to change the scripts that setup the vector databases with these embeddings? Like https://github.com/GoogleCloudPlatform/cymbal-air-toolbox-demo/blob/7a405212542aac30c130250946dfb47552410fce/data/run_generate_embeddings.py#L22

tools.yaml
emagda97: nice catch! swapped it for it.

data/run_database_init.py
emagda97: We removed it because with MCP Toolbox (embeddedBy), embedding generation is handled by the toolbox on the client side and passed in as a raw vector ($1::vector). The database itself no longer calls embedding(...) in SQL, so google_ml_integration is no longer needed (only the vector extension is required).

anubhav756 (approved):
Thanks for the PR!

Just wanting to know if we have a way to test these changes and verify?

tools.yaml
anubhav756: I still see the old model. Am I missing something?

anubhav756 (approved):
nit: I still saw the old model. Am I missing something? :)

